Issue #21, opened 2016-03-23
bazel error: export_half_plus_two.py, line 66, SyntaxError: Missing parentheses in call to 'print'


seetea:
**ERROR:**  
`
ERROR: /home/seetea/tensorflow-serving-latest/tensorflow_serving/session_bundle/example/BUILD:34:1: Executing genrule //tensorflow_serving/session_bundle/example:half_plus_two failed: bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
  File "/home/seetea/.cache/bazel/_bazel_seetea/8d38ebe87742a96afdc093845beacfa5/tensorflow-serving-latest/bazel-out/host/bin/tensorflow_serving/session_bundle/example/export_half_plus_two.runfiles/tensorflow_serving/session_bundle/example/export_half_plus_two.py", line 66
    print "copying asset files to: %s" % export_path
                                     ^
SyntaxError: Missing parentheses in call to 'print'
`

**Environment:** CentOS 7 + Python 3.4

**Description:** When I use "bazel build tensorflow_serving/..." to build the tensorflow_serving source codes under python 3.4, I met the above error. 

**Solution:**
I made the following modifications to make the codes work for both python 2.x and python 3.x:

1. In "serving/tensorflow_serving/session_bundle/example/export_half_plus_two.py", add
`from __future__ import print_function`
and change
`print "copying asset files to: %s" % export_path`
to 
`print ("copying asset files to: %s" % export_path)`

2. In "serving/tensorflow_serving/session_bundle/exporter.py", 
```
import six
.....
#for name, tensor in name_tensor_map.iteritems():
for name, tensor in six.iteritems(name_tensor_map):
......
#for signature_name, signature in named_graph_signatures.iteritems():
for signature_name, signature in six.iteritems(named_graph_signatures):
.......
```

kirilg:
Thanks for reporting and the suggested fix! Did you want to send that out as a PR or would you prefer for us to make those changes?

kirilg:
Fix PR was merged, closing issue. Thanks!

PR #26, opened 2016-03-29
Fix issue #21,support both python 2.7 and python 3.4

seetea:
Fix issue #21, to make "serving/tensorflow_serving/session_bundle/example/export_half_plus_two.py" and "serving/tensorflow_serving/session_bundle/exporter.py" support both python 2.7 and python 3.4.


seetea:
I have signed the Individual CLA. Thanks.

PR #57, opened 2016-04-25
Implemented batching for Inception example.

eriophora:
Implemented batching for the inception example with TensorFlow Serving. This was accomplished by performing client-side preprocessing, using PIL. It appears, however, that this requires a newer version of PIL than is provided with TensorFlow, and it has to be imported before TensorFlow to avoid bizarre filestream issues with PIL. 

Additionally, three types of preprocessing were implemented:
- Resize (the default -- and original way)
- Crop (the image is resized to preserve the maximum amount of image possible then cropped to the correct size)
- Pad & Resize (the image is symmetrically padded to the correct aspect ratio then resized)

See issue #38


eriophora:
I signed it!

vinuraja:
@fangweili could you please take a look at this ?

tensorflow_serving/example/inception_inference.cc
```diff
@@ -208,10 +212,9 @@ InceptionServiceImpl::InceptionServiceImpl(
   // the numbers are extremely performance critical and should be tuned based
   // specific graph structure and usage.
   tensorflow::serving::StreamingBatchScheduler<Task>::Options scheduler_options;
-  scheduler_options.thread_pool_name = "inception_service_batch_threads";
-  // TODO(27776734): Current exported model supports only batch_size=1
-  // See inception_export.py for details.
-  scheduler_options.max_batch_size = 1;
+  scheduler_options.batch_timeout_micros = 1000 * 1000;  // 1 second
+  scheduler_options.num_batch_threads = 4;
+  scheduler_options.max_batch_size = 22;
```
chrisolston: Maybe add a comment on how you selected 22 as the max batch size?

Similar for num_batch_threads: are you aiming to match the server machine's number of GPU devices or ??

eriophora:
Actually, it was simply the result of testing. This was tested on an Amazon EC2 GPU g2.2xlarge. The original number of images, 32, is too large for these machines and they will repeatedly emit those errors--I can't remember exactly what they are, orphaned or ejected memory or somesuch--and then crash.

The number of threads is somewhat arbitrary, but having at least 4 helps to avoid the "no available batches" as it appears that TF Serving will start evaluating a batch that's not full but that has timed out even if the compute resource is not yet free. 

However, to get the whole thing working on EC2 required quite a bit of modification -- and as such this could certainly benefit from some testing on a 'normal' set up.

sukritiramesh:
@fangweili Can you please take a look at the PR? Thanks!

fangweili:
@eriophora Sorry for the late response. This looks really good. I wonder if you can fork files into inception_batching_* instead and submit as a separate example. We would like to keep the original inception example as simple as possible, for tutorial purpose. You can update the TODO in inception_export.py to point to this new example. Thanks a lot!

tensorflow_serving/example/inception_inference.proto
```diff
@@ -1,12 +1,11 @@
...
-
 syntax = "proto3";
 
 package tensorflow.serving;
 
 message InceptionRequest {
-  // JPEG encoded stream of the image to be classified.
-  bytes jpeg_encoded = 1;
+  // Image data as a flattened array of floats
+  repeated float image_data = 1 [packed = true];
```
rcompton: is this necessary for batching?

chrisolston:
This PR has been sitting for a while now. I'm going to close it due to inactivity. Feel free to reopen.

PR #64, opened 2016-05-03
Increases //tensorflow_serving/core:basic_manager_test test size

damienmg:
This test is timeout flaky on ci.bazel.io, see:
  http://ci.bazel.io/job/TensorFlow_Serving/81/BAZEL_VERSION=HEAD,PLATFORM_NAME=linux-x86_64/console
  http://ci.bazel.io/job/TensorFlow_Serving/84/BAZEL_VERSION=HEAD,PLATFORM_NAME=linux-x86_64/console

### DIFF
--- a/tensorflow_serving/core/BUILD
+++ b/tensorflow_serving/core/BUILD
@@ -454,7 +454,7 @@ cc_library(
 
 cc_test(
     name = "basic_manager_test",
-    size = "small",
+    size = "medium",
     srcs = ["basic_manager_test.cc"],
     deps = [
         "@tf//tensorflow/core:lib",


vinuraja:
Thanks for patch @damienmg ! When I look at the logs it says the output for that test can be found at 
`/home/ci/.cache/bazel/_bazel_ci/e409f7152fa45ddf3337f1660b5273b9/linux-x86_64/bazel-out/local-fastbuild/testlogs/tensorflow_serving/core/basic_manager_test/test.log`

But how do I look at these logs ?

damienmg:
The log is not archived, so it get replaced by newer version. However the console output is archived as well as the XML file. Unfortunately the XML file is not created when timing out (Anyway the default one does not contains the output). However we use --test_output=errors so the test output is printed in the console output: http://ci.bazel.io/job/TensorFlow_Serving/84/BAZEL_VERSION=HEAD,PLATFORM_NAME=linux-x86_64/console

vinuraja:
Thanks! It turned out to be a bug, which we have fixed upstream. The next push should have the fix.

PR #63, opened 2016-04-30
fixed a bit of confusing documentation

eriophora:
The documentation incorrectly states that `num_batch_threads` governs the number of batches worth of tasks that may be enqueued.

PR #67, opened 2016-05-06
Fixed documentation in the batches

eriophora:
The old PR was #63, but since I pushed from multiple computers apparently this screwed up the CLA. I also had to square the commits down into one.


eriophora:
Oh my god googlebot are you serious!? %(*&@#$!!

PR #101, opened 2016-06-08
Remove `grte/v4` from examples

rcompton:
While going through the examples I came across a few instances of `#!/usr/grte/v4/bin/python2.7`. `grte/v4` is an internal Google tool, right? This pull request sets a standard shebang line.

### DIFF
--- a/tensorflow_serving/example/inception_client.py
+++ b/tensorflow_serving/example/inception_client.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 
 """Send JPEG image to inception_inference server for classification.
 """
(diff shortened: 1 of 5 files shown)


rcompton:
`I signed it!`

tensorflow_serving/example/inception_client.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
-#!/usr/grte/v4/bin/python2.7
+#!/usr/bin/env python
```
kirilg: Did you mean to say
#!/usr/bin/env python2.7
in all of these files?

We should hopefully remove this line entirely soon, but gRPC currently doesn't support python3 so it would probably be nice to keep 2.7 as the default here.

rcompton: 2.7 makes more sense. I'll change it.

kirilg:
@tensorflow-jenkins test this please

kirilg:
Thanks @rcompton, looks good, just running tests now.

Issue #139, opened 2016-07-25
Several tests are broken on ci.bazel.io


damienmg:
See http://ci.bazel.io/job/TensorFlow_Serving/BAZEL_VERSION=latest,PLATFORM_NAME=linux-x86_64/214/console

Happens with both Bazel @ HEAD and 0.3.0

Failing tests:
```
//tensorflow_serving/batching:batching_session_test                   NO STATUS
//tensorflow_serving/servables/tensorflow:session_bundle_factory_test NO STATUS
//tensorflow_serving/servables/tensorflow:session_bundle_source_adapter_test NO STATUS
//tensorflow_serving/session_bundle:session_bundle_py_test            NO STATUS
//tensorflow_serving/session_bundle:exporter_test                        FAILED in 3.6s
  /home/ci/.cache/bazel/_bazel_ci/3765d23dccdc53317a8257c675fcb26d/execroot/linux-x86_64/bazel-out/local-fastbuild/testlogs/tensorflow_serving/session_bundle/exporter_test/test.log
//tensorflow_serving/session_bundle:gc_test                              FAILED in 3.6s
  /home/ci/.cache/bazel/_bazel_ci/3765d23dccdc53317a8257c675fcb26d/execroot/linux-x86_64/bazel-out/local-fastbuild/testlogs/tensorflow_serving/session_bundle/gc_test/test.log
//tensorflow_serving/util:fast_read_dynamic_ptr_benchmark                FAILED in 16.5s
  /home/ci/.cache/bazel/_bazel_ci/3765d23dccdc53317a8257c675fcb26d/execroot/linux-x86_64/bazel-out/local-

```

And interesting part of the log:
```
ERROR: /home/ci/.cache/bazel/_bazel_ci/3765d23dccdc53317a8257c675fcb26d/external/org_tensorflow/tensorflow/contrib/session_bundle/example/BUILD:38:1: Executing genrule @org_tensorflow//tensorflow/contrib/session_bundle/example:half_plus_two failed: bash failed: error executing command 
  (cd /home/ci/.cache/bazel/_bazel_ci/3765d23dccdc53317a8257c675fcb26d/execroot/linux-x86_64 && \
  exec env - \
    PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
[...]

Issue #133, opened 2016-07-20
simple_servers_test timeout in bazel test tensorflow_serving/...


chenleiwhu:
I used serving stack in docker, Mem is ~16GB

kirilg:
Is it consistently timing out? What if you run `bazel test tensorflow_serving/servables/tensorflow:simple_servers_test` a few times, does it always fail?

Could you please provide some additional test logs that may be relevant?

chenleiwhu:
it failed consistently

the failure log is below
```
> exec ${PAGER:-/usr/bin/less} "$0" || exit 1
> -----------------------------------------------------------------------------
> Running main() from test_main.cc
> [==========] Running 1 test from 1 test case.
> [----------] Global test environment set-up.
> [----------] 1 test from SimpleServersTest
> [ RUN      ] SimpleServersTest.Basic
> I tensorflow_serving/core/basic_manager.cc:190] Using InlineExecutor for BasicManager.
> E tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc:133] FileSystemStoragePathSource encountered a file-system access error: Could not find base path: /root/.cache/bazel/_bazel_root/f8d1071c69ea316497c31e40fe01608c/serving/bazel-out/local_linux-fastbuild/bin/tensorflow_serving/servables/tensorflow/simple_servers_test.runfiles/tf_serving/external/org_tensorflow/tensorflow/contrib/session_bundle/example/half_plus_two
> E tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc:133] FileSystemStoragePathSource encountered a file-system access error: Could not find base path: /root/.cache/bazel/_bazel_root/f8d1071c69ea316497c31e40fe01608c/serving/bazel-out/local_linux-fastbuild/bin/tensorflow_serving/servables/tensorflow/simple_servers_test.runfiles/tf_serving/external/org_tensorflow/tensorflow/contrib/session_bundle/example/half_plus_two
```



```
root@f9c73b86092b:/serving# bazel test tensorflow_serving/servables/tensorflow:simple_servers_test
INFO: Reading 'startup' options from /root/.bazelrc: --batch
WARNING: Sandboxed execution is not supported on your system and thus hermeticity of actions cannot be guaranteed. See h
ttp://bazel.io/docs/bazel-user-manual.html#sandboxing for more information. You can turn off this warning via --ignore_u
nsupported_sandboxing.
INFO: Found 1 test target...
[...]

PR #169, opened 2016-08-31
Fixes missing mock lib (#139) and failed unit tests (#133)

pythonner:
From a newly built Docker:

`Executed 46 out of 46 tests: 46 tests pass.`


pythonner:
I signed it!

tensorflow_serving/tools/docker/Dockerfile.devel
```diff
@@ -28,7 +28,7 @@ RUN curl -fSsL -O https://bootstrap.pypa.io/get-pip.py && \
 
 # Set up grpc
 
-RUN pip install enum34 futures six && \
+RUN pip install enum34 futures mock six && \
     pip install --pre protobuf>=3.0.0a3 && \
```
promiseofcake: While we are here, would it be possible to add `'protobuf>=3.0.0a3'`, it seems without it will touch a file on the root filesystem named `=3.0.0a3`, and I am not sure what version of protobuf is truly installed due to that.

pythonner: Ok, I confirmed it builds well, the file `=3.0.0a3` is not there anymore and all tests pass.
`Executed 46 out of 46 tests: 46 tests pass.`

tensorflow_serving/tools/docker/Dockerfile.devel
```diff
@@ -56,7 +56,7 @@ RUN echo "build --spawn_strategy=standalone --genrule_strategy=standalone" \
     >>/root/.bazelrc
 ENV BAZELRC /root/.bazelrc
 # Install the most recent bazel release.
-ENV BAZEL_VERSION 0.2.0
+ENV BAZEL_VERSION 0.3.1
```
kirilg: I missed that you had this PR, but I already updated the BAZEL_VERSION to 0.3.0 (What TF uses) and added a WORKSPACE file check so that you get a useful error message right away.

You can revert that change, but keep the mock change above and we'll merge. Thanks!

PR #171, opened 2016-09-08
delete unnecessary importing itself in beta mnistservice functions

AppleHolic:
When I was [following the tutorials](https://tensorflow.github.io/serving/serving_basic), I found unnecessary importing codes.

So, I test the codes after fixing it, I cannot found any problems.

If any reasons to code the twice of the importing itself, can you tell me about the reasons?

If no problems with delete code lines, I hope this PR will be merged !

Thanks for making tutorials.


AppleHolic:
Ah.. this code is auto-generated by proto. I got it and close the PR.

PR #173, opened 2016-09-09
Instead of appending all of each image values, fix it to extend items…

AppleHolic:
… at once

Where I follow the example codes in mnist_client.py, I found more efficient way to append image data to request.

As fixing code like below, about 2x faster in the loop. (350ms -> 190ms, In the loop that starts from 91th line on the jupyter notebook)

And the user time of entire built source is decreased 4.5 sec to 3.2 sec.

So I suggest fixing it!

### DIFF
--- a/tensorflow_serving/example/mnist_client.py
+++ b/tensorflow_serving/example/mnist_client.py
@@ -91,8 +91,7 @@ def do_inference(hostport, work_dir, concurrency, num_tests):
   for _ in range(num_tests):
     request = mnist_inference_pb2.MnistRequest()
     image, label = test_data_set.next_batch(1)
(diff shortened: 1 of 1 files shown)


kirilg:
Thanks for the PR! We've recently changed all of the examples in tensorflow_serving/example (including mnist) to use the standard [model_server](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/main.cc) instead of custom servers for each example. As a result, the request type is now [PredictRequest](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/predict.proto), so this change is no longer compatible with the latest client at head.

Please give the new client a try and let us know if you encounter any issues.

PR #184, opened 2016-09-20
Ignore built /bazel-tensorflow_serving file

promiseofcake:
`/bazel-tensorflow_serving` is built and stored in the repository root, this will prevent it from being tracked.


kirilg:
Jenkins, test this please

Issue #121, opened 2016-07-09
TF serving as standalone app


arielscarpinelli:
The common use case for TF Serving should be running inference over an exported TF model. As such seems to me a bit useless to repeat always the same boilerplate code to create a new server, load the exported graph, map the input request protobuf to the placeholdes by the means of the signature, call inference and then map back the output using signature to a result protobuf. 

Can't this be simply a generic process run thru an standalone executable that loads models thru configuration, no need build your own? I believe you could define a generic enough service protobuf that could accomodate any kind of input and even validate before trying to run inference by maybe adding a bit more information in the signatures.

I envision this as an app that you could get to install in your server thru `sudo apt-get install tensorflow-serving`

kirilg:
Good points. I'll mention first that the intent of TensorFlow Serving is to create a set of libraries for building ML servers in a flexible way by using whatever frontend/plugins you wish. In our examples we used gRPC with a custom proto service (to not give the illusion that it's a stable reusable API - it's just an example) + Tensorflow plugins. However, you could envision others not wanting to use gRPC/protos and specifying their own frontend.

With that said, your points are valid. We could still provide a certain gRPC implementation with a stable proto API which may cover the majority of use cases out there, and that's something we're definitely thinking about. We didn't want to rush an API, but stayed tuned.

arielscarpinelli:
Cool! If the desire is to keep it open and not tie it to gRPC, a "plain old JSON REST API" support would be great too :-D

arielscarpinelli:
I was trying to support REST myself but couldn't find any google supported, blaze oriented, dropin libraries for building HTTP based webservices on C++

chenleiwhu:
It's pretty cool to see you already have the plan to do it. @arielscarpinelli you can use any wrapper service framework like node.js or tornado in python.

arielscarpinelli:
I already done it with Python. The spirit of the request is to avoid having
to;
- code the server boilerplate in C++
- avoid the need for two servers

kirilg:
We recently released the [model_server](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/main.cc) which addresses your two points above (reusable server + [server_core](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/server_core.h) that eliminates a lot of the boilerplate in C++).

[...]

PR #192, opened 2016-09-27
Modify server_core visibility for external server components

promiseofcake:
Update visibility on `server_core`, `serving_session` and `predict_impl` to aid in building a custom server outside of the tensorflow_serving package.

Example directory structure:
```
tensorflow_serving/...
tensorflow/...
tf_models/...
promiseofcake/server/server.cc
```

@chrisolston, would all of these declarations be valid, or is `PredictionService` more of an internal definition?

Based upon discussion here: https://github.com/tensorflow/serving/issues/121, https://github.com/tensorflow/serving/issues/121#issuecomment-250011716

### DIFF
--- a/tensorflow_serving/servables/tensorflow/BUILD
+++ b/tensorflow_serving/servables/tensorflow/BUILD
@@ -199,6 +199,9 @@ cc_library(
     name = "serving_session",
     srcs = ["serving_session.cc"],
     hdrs = ["serving_session.h"],
+    visibility = [
+        "//visibility:public",
+    ],
     deps = [
         "@org_tensorflow//tensorflow/core:core_cpu",
         "@org_tensorflow//tensorflow/core:lib",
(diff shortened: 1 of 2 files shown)


tensorflow_serving/servables/tensorflow/BUILD
```diff
@@ -199,6 +199,9 @@ cc_library(
     name = "serving_session",
     srcs = ["serving_session.cc"],
     hdrs = ["serving_session.h"],
+    visibility = [
+        "//visibility:public",
```
chrisolston: I'm not sure we want to open up the visibility of serving_session and predict_impl. Do you need those to be open? If so, can you describe your use-case? Thanks.

promiseofcake:
Thanks @chrisolston, the use case I am trying to solve for would be building my own version of the standard model server roughly attempting to follow along here: https://tensorflow.github.io/serving/serving_advanced. The main motivation being I would like to define an IDL to speak from my external applications to the serving service over gRPC which would then perform some business logic on our behalf before making a prediction request to the loaded model.

I _could_ re-purpose the old inception_inference example and build off of that, but I wanted to take advantage of the dynamic model loading available in the main.cc server. 

Perhaps I have over-extended what I need, and servercore should be enough for that?

chrisolston:
I *think* server_core would be enough for what you want to do. How about we start with just changing the visibility of server_core, and then if needed revisit the others.

promiseofcake:
Sounds good @chrisolston, will update. Thank you.

promiseofcake:
Implemented upstream! Thanks.

PR #214, opened 2016-10-22
add command line option for file system poll wait (seconds)

waichee:
Add an option on the command line for changing the default 30s file system poll wait interval.

### DIFF
--- a/tensorflow_serving/model_servers/main.cc
+++ b/tensorflow_serving/model_servers/main.cc
@@ -185,13 +185,13 @@ int main(int argc, char** argv) {
   tensorflow::int32 port = 8500;
   bool enable_batching = false;
   tensorflow::string model_name = "default";
-  tensorflow::int32 file_system_polling_interval = -1;
+  tensorflow::int32 file_system_poll_wait_seconds = 1;
   tensorflow::string model_base_path;
   std::vector<tensorflow::Flag> flag_list = {
       tensorflow::Flag("port", &port, "port to listen on"),
       tensorflow::Flag("enable_batching", &enable_batching, "enable batching"),
       tensorflow::Flag("model_name", &model_name, "name of model"),
-      tensorflow::Flag("file_system_polling_interval", &file_system_polling_interval, "interval in seconds between each poll of the file system for new model"),
+      tensorflow::Flag("file_system_poll_wait_seconds", &file_system_poll_wait_seconds, "interval in seconds between each poll of the file system for new model version"),
       tensorflow::Flag("model_base_path", &model_base_path,
                        "path to export (required)")};
   string usage = tensorflow::Flags::Usage(argv[0], flag_list);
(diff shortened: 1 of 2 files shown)


waichee:
I signed it!

waichee:
Jenkins, test this please

waichee:
@tensorflow-jenkins test this please

tensorflow_serving/model_servers/main.cc
```diff
@@ -219,6 +221,11 @@ int main(int argc, char** argv) {
   core_config.aspired_version_policy =
       std::unique_ptr<AspiredVersionPolicy>(new EagerLoadPolicy);
 
+  // Override default file polling interval
+  if (file_system_polling_interval > 0) {
+    core_config.file_system_poll_wait_seconds = file_system_polling_interval;
```
chrisolston: I'd prefer to give the flag the same name as the config variable, to minimize confusion. (Even if you may not like the name as much, I feel it's better to keep it consistent across the layers of the system.)

tensorflow_serving/model_servers/tensorflow_model_server_test.py
```diff
@@ -68,13 +68,14 @@ def TerminateProcs(self):
     if self.server_proc is not None:
       self.server_proc.terminate()
 
-  def RunServer(self, port, model_name, model_path):
+  def RunServer(self, port, model_name, model_path, file_system_polling_interval):
```
waichee: Sure, ill leave the test case as only for the cases where the flag is omitted

waichee:
@chrisolston  
- updated the command line option name to matching existing name
- revert test change

tensorflow_serving/model_servers/main.cc
```diff
@@ -185,11 +185,13 @@ int main(int argc, char** argv) {
   tensorflow::int32 port = 8500;
   bool enable_batching = false;
   tensorflow::string model_name = "default";
+  tensorflow::int32 file_system_poll_wait_seconds = -1;
```
chrisolston: I don't think -1 is (or should be) a valid value. Can we make the default 1 instead of -1 please.

waichee:
sure updated from -1 to 1 and fix up description text

tensorflow_serving/model_servers/main.cc
```diff
@@ -219,6 +221,11 @@ int main(int argc, char** argv) {
   core_config.aspired_version_policy =
       std::unique_ptr<AspiredVersionPolicy>(new EagerLoadPolicy);
 
+  // Override default file polling interval
+  if (file_system_poll_wait_seconds > 1) {
```
chrisolston: Please drop the if-statement. (Setting it to 0 is valid, and the FileSystemStoragePathSource does (or at least should) do validation of the argument.)

waichee: sure, if statement removed.

Issue #197, opened 2016-09-29
serving multiple version of the same model


waichee:
We are looking at serving multiple versions of the same model (i.e. `000001/` and `000002/`). The default behavior of `AspireVersionManager` is to eagerly load the newest version and unload all version before the latest version. What would be the recommended way of serving 2 versions of the same models in TS serving say for e.g. for AB testing purpose? The options that I could see are
- modify the eager unload policy to handle this.
- change to the Caching manager rather than AspireVersion Manager?
- change `FileSystemStoragePathSource` to aspire latest version and version before it?

Also if we do change the graph structure of the model (and not just updating the weights) in the new version, would Tensorflow serving be able to handle this or do we need to declare the new version as a different model (with different name in the model spec)?

chrisolston:
The right way to do it is to define a new Source that aspires multiple versions at the same time.

And no, TF-Serving doesn't care if the model structure varies across versions.

waichee:
cool thanks good to know @chrisolston

sukritiramesh:
Hi @waichee, it looks like this issue is now resolved. Closing for now; but please feel free to reopen if you have further questions.

PR #217, opened 2016-10-23
Add options to serve multiple versions of model

waichee:
Modify `FileSystemStoragePathSource` to emit all versions rather than just the latest version if the following environment variable is set `TF_SERVING_MULTIPLE_VERSIONS` is set to true.

This does not change the default behavior of only serving the latest version.

This is discussed in the following Github issue :
https://github.com/tensorflow/serving/issues/197

### DIFF
--- a/tensorflow_serving/g3doc/serving_advanced.md
+++ b/tensorflow_serving/g3doc/serving_advanced.md
@@ -142,9 +142,8 @@ With all these, `ServerCore` internally does the following:
   `AspiredVersionsManager` that manages all such `Loader` instances created by
   the `SessionBundleSourceAdapter`.
 
-Whenever a new version is available, this `AspiredVersionsManager` always
-unloads the old version and replaces it with the new one. If you want to
-start customizing, you are encouraged to understand the components that it
+Whenever a new version is available, this `AspiredVersionsManager` loads the new version, and, 
+under its default behavior, unloads the old one. If you want to start customizing, you are encouraged to understand the components that it
 creates internally, and how to configure them.
 
 It is worth mentioning that TensorFlow Serving is designed from scratch to be
--- a/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
+++ b/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
@@ -97,13 +97,13 @@ bool ParseVersionNumber(const string& version_path, int64 *version_number) {
 // Update the servable data to include
 // all the model versions found in the base path as aspired versions.
 // The argument 'children' represents a list of base-path children from the file system.
-// Returns true if valid servable path is found, otherwise returns false.
+// Returns true if one or more valid servable version paths are found, otherwise returns false.
 bool AspireAllVersions(
     const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
     const std::vector<string>& children,   // represents a list of base-path children from the file system.
     std::vector<ServableData<StoragePath>>* versions) {
 
-  bool version_found = false;
+  bool at_least_one_version_found = false;
   for (const string& child : children) {
     // Identify all the versions, among children that can be interpreted as
     // version numbers.
@@ -111,18 +111,18 @@ bool AspireAllVersions(
     if (ParseVersionNumber(child, &version_number)) {
       // Emit all the aspired-versions data.
       AspireVersion(servable, child, version_number, versions);
-      version_found = true;
+      at_least_one_version_found = true;
     }
   }
 
-  return version_found;
+  return at_least_one_version_found;
 }
 
 // Update the servable data to include
 // the latest version found in the base path as aspired version.
 // The argument 'children' represents a list of base-path children from the file system.
 // Returns true if valid servable path is found, otherwise returns false.
-bool AspireLatestVersions(
+bool AspireLatestVersion(
     const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
     const std::vector<string>& children,
     std::vector<ServableData<StoragePath>>* versions) {
@@ -131,7 +131,7 @@ bool AspireLatestVersions(
   // version numbers.
   int latest_version_child_index;
   int64 latest_version_number;
-  bool version_found = false;
+  bool at_least_one_version_found = false;
   for (int i = 0; i < children.size(); ++i) {
     int64 version_number;
     if (!ParseVersionNumber(children[i], &version_number)) {
@@ -139,19 +139,19 @@ bool AspireLatestVersions(
     }
 
     // Check if this is the largest version number
-    if (!version_found || latest_version_number < version_number) {
+    if (!at_least_one_version_found || latest_version_number < version_number) {
       latest_version_child_index = i;
       latest_version_number = version_number;
     }
-    version_found = true;
+    at_least_one_version_found = true;
   }
 
   // Emit the latest aspired version.
-  if (version_found) {
+  if (at_least_one_version_found) {
     AspireVersion(servable, children[latest_version_child_index], latest_version_number, versions);
   }
 
-  return version_found;
+  return at_least_one_version_found;
 }
 
 // Like PollFileSystemForConfig(), but for a single servable.
@@ -183,17 +183,17 @@ Status PollFileSystemForServable(
   children.clear();
   children.insert(children.begin(), real_children.begin(), real_children.end());
 
-  bool version_found = false;
+  bool at_least_one_version_found = false;
   switch (servable.version_policy()) {
     case FileSystemStoragePathSourceConfig::LATEST_VERSION:
-      version_found = AspireLatestVersions(servable, children, versions);
+      at_least_one_version_found = AspireLatestVersion(servable, children, versions);
       break;
     case FileSystemStoragePathSourceConfig::ALL_VERSIONS:
-      version_found = AspireAllVersions(servable, children, versions);
+      at_least_one_version_found = AspireAllVersions(servable, children, versions);
       break;
   }
 
-  if (!version_found) {
+  if (!at_least_one_version_found) {
     LOG(WARNING) << "No versions of servable " << servable.servable_name()
                  << " found under base path " << servable.base_path();
   }
--- a/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
+++ b/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
@@ -76,6 +76,84 @@ std::set<string> GetDeletedServables(
   return deleted_servables;
 }
 
+//Add a new ServableData for the model version to the the vector of versions to aspire
+void AspireVersion(
+    const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
+    const string& version_relative_path,
+    const int version_number,
+    std::vector<ServableData<StoragePath>>* versions) {
+    const ServableId servable_id = {servable.servable_name(), version_number};
+    const string full_path =
+        io::JoinPath(servable.base_path(), version_relative_path);
+    versions->emplace_back(ServableData<StoragePath>(servable_id, full_path));
+}
+
+// Attempts to parse the version path as integer, upon success writes the result to 'version_number' and returns true.
+// Otherwise returns false when fail to parse the version path.
+bool ParseVersionNumber(const string& version_path, int64 *version_number) {
+  return strings::safe_strto64(version_path.c_str(), version_number);
+}
+
+// Update the servable data to include
+// all the model versions found in the base path as aspired versions.
+// The argument 'children' represents a list of base-path children from the file system.
+// Returns true if valid servable path is found, otherwise returns false.
+bool AspireAllVersions(
+    const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
+    const std::vector<string>& children,   // represents a list of base-path children from the file system.
+    std::vector<ServableData<StoragePath>>* versions) {
+
+  bool version_found = false;
+  for (const string& child : children) {
+    // Identify all the versions, among children that can be interpreted as
+    // version numbers.
+    int64 version_number;
+    if (ParseVersionNumber(child, &version_number)) {
+      // Emit all the aspired-versions data.
+      AspireVersion(servable, child, version_number, versions);
+      version_found = true;
+    }
+  }
+
+  return version_found;
+}
+
+// Update the servable data to include
+// the latest version found in the base path as aspired version.
+// The argument 'children' represents a list of base-path children from the file system.
+// Returns true if valid servable path is found, otherwise returns false.
+bool AspireLatestVersions(
+    const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
+    const std::vector<string>& children,
+    std::vector<ServableData<StoragePath>>* versions) {
+
+  // Identify the latest version, among children that can be interpreted as
+  // version numbers.
+  int latest_version_child_index;
+  int64 latest_version_number;
+  bool version_found = false;
+  for (int i = 0; i < children.size(); ++i) {
+    int64 version_number;
+    if (!ParseVersionNumber(children[i], &version_number)) {
+      continue;
+    }
+
+    // Check if this is the largest version number
+    if (!version_found || latest_version_number < version_number) {
+      latest_version_child_index = i;
+      latest_version_number = version_number;
+    }
+    version_found = true;
+  }
+
+  // Emit the latest aspired version.
+  if (version_found) {
+    AspireVersion(servable, children[latest_version_child_index], latest_version_number, versions);
+  }
+
+  return version_found;
+}
+
 // Like PollFileSystemForConfig(), but for a single servable.
 Status PollFileSystemForServable(
     const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
@@ -105,30 +183,17 @@ Status PollFileSystemForServable(
   children.clear();
   children.insert(children.begin(), real_children.begin(), real_children.end());
 
-  // Identify the latest version, among children that can be interpreted as
-  // version numbers.
-  int latest_version_child = -1;
-  int64 latest_version;
-  for (int i = 0; i < children.size(); ++i) {
-    const string& child = children[i];
-    int64 child_version_num;
-    if (!strings::safe_strto64(child.c_str(), &child_version_num)) {
-      continue;
-    }
-
-    if (latest_version_child < 0 || latest_version < child_version_num) {
-      latest_version_child = i;
-      latest_version = child_version_num;
-    }
+  bool version_found = false;
+  switch (servable.version_policy()) {
+    case FileSystemStoragePathSourceConfig::LATEST_VERSION:
+      version_found = AspireLatestVersions(servable, children, versions);
+      break;
+    case FileSystemStoragePathSourceConfig::ALL_VERSIONS:
+      version_found = AspireAllVersions(servable, children, versions);
+      break;
   }
 
-  // Emit the aspired-versions data.
-  if (latest_version_child >= 0) {
-    const ServableId servable_id = {servable.servable_name(), latest_version};
-    const string full_path =
-        io::JoinPath(servable.base_path(), children[latest_version_child]);
-    versions->emplace_back(ServableData<StoragePath>(servable_id, full_path));
-  } else {
+  if (!version_found) {
     LOG(WARNING) << "No versions of servable " << servable.servable_name()
                  << " found under base path " << servable.base_path();
   }
--- a/tensorflow_serving/sources/storage_path/file_system_storage_path_source.proto
+++ b/tensorflow_serving/sources/storage_path/file_system_storage_path_source.proto
@@ -4,6 +4,17 @@ package tensorflow.serving;
 
 // Config proto for FileSystemStoragePathSource.
 message FileSystemStoragePathSourceConfig {
+
+  // The policy to define how many versions of the servable should be
+  // served at the same time.
+  enum VersionPolicy {
+    // Only serve the latest version that exists in the base path.
+    // This is the default behavior.
+    LATEST_VERSION = 0;
+    // Serves all the versions that exist in the base path.
+    ALL_VERSIONS = 1;
+  }
+
   // A servable name and base path to look for versions of the servable.
   message ServableToMonitor {
     // The servable name to supply in aspired-versions callback calls. Child
@@ -12,6 +23,10 @@ message FileSystemStoragePathSourceConfig {
 
     // The path to monitor, i.e. look for child paths of the form base_path/123.
     string base_path = 2;
+
+    // The policy to determines the number of versions of the servable to be served at the same time.
+    VersionPolicy version_policy = 3;
+
   };
 
   // The servables to monitor for new versions, and aspire.
--- a/tensorflow_serving/sources/storage_path/file_system_storage_path_source_test.cc
+++ b/tensorflow_serving/sources/storage_path/file_system_storage_path_source_test.cc
@@ -171,6 +171,39 @@ TEST(FileSystemStoragePathSourceTest, MultipleVersions) {
                    .PollFileSystemAndInvokeCallback());
 }
 
+TEST(FileSystemStoragePathSourceTest, MultipleVersionsAtTheSameTime) {
+  const string base_path = io::JoinPath(testing::TmpDir(), "MultipleVersionsAtTheSameTime");
+  TF_ASSERT_OK(Env::Default()->CreateDir(base_path));
+  TF_ASSERT_OK(Env::Default()->CreateDir(
+      io::JoinPath(base_path, "non_numerical_child")));
+  TF_ASSERT_OK(Env::Default()->CreateDir(io::JoinPath(base_path, "42")));
+  TF_ASSERT_OK(Env::Default()->CreateDir(io::JoinPath(base_path, "17")));
+
+  string all_version_policy = std::to_string(FileSystemStoragePathSourceConfig::ALL_VERSIONS);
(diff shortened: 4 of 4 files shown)


chrisolston:
Hi waichee,

Thanks for proposing this change.

We try hard to avoid global variables/flags, even for configuration. The FileSystemStoragePathSourceConfig proto would be the appropriate place to put that option. Probably the cleanest approach would be to define some sort of VersionsPolicy enum and add it to ServableToMonitor, so for each servable you would specify the path and the version policy (just the largest version#, or all versions). (That way, if in the future we want more options e.g. the two largest versions we could add those to the enum.) Please be sure that the default (i.e. the 0th enum value) is equivalent to the current semantics (just the largest version#).

-Chris

waichee:
Thanks for reviewing @chrisolston 
I agreed consistency is important.  Ill revise the change to use custom version policy class instead of environment variables as toggle to determine the number of versions served

chrisolston:
Sounds good, thanks.

tensorflow_serving/g3doc/serving_advanced.md
```diff
@@ -143,7 +143,17 @@ With all these, `ServerCore` internally does the following:
...
 Whenever a new version is available, this `AspiredVersionsManager` always
-unloads the old version and replaces it with the new one. If you want to
+unloads the old version and replaces it with the new one. The default behavior means that
+there will only be one version of the model served at a given time (and that version will be the 
+latest version). The latest version is determined from the model folder name which should be numeric.
+If you would like to serve all the versions available in `model_base_path`, you can set the following environment 
+variable before compiling Tensorflow Serving from source.
+
+~~~
+export TF_SERVING_MULTIPLE_VERSIONS=1
```
waichee: I think for those that are starting out to play with TS serving, they most likely do not need this feature. But once they get to productionizing their model serving pipeline, i believe versioning and A/B testing is generally becomes fairly important. 

Is there other places where the features of TS serving is documented besides in the tutorials? Or usually it's expected for users to read the codebase ?

chrisolston: The tutorials are not intended to be exhaustive, but merely to get users started with basic functionality. I think expecting folks hunting for advanced features to read the config proto definitions (not the codebase) is reasonable. The config proto definitions are generally well documented, and serve as the primary/only source of documentation about advanced features.

waichee:
@chrisolston  updated to use enum of version policy within the proto files for determining which the versions to aspired/emit. Would appreciate it if  you can review the updated changes.

tensorflow_serving/g3doc/serving_advanced.md
waichee: i've removed the description on how to enable multiple versions.

The config protos has good comments on them. However it would be good if there's API docs or more details for advanced features. This would help users that intend to use tensorflow serving in production systems.

Perhaps along the line of dataflow docs or spark docs?
https://cloud.google.com/dataflow/docs/

tensorflow_serving/sources/storage_path/file_system_storage_path_source.proto
```diff
@@ -4,6 +4,16 @@ package tensorflow.serving;
...
 // Config proto for FileSystemStoragePathSource.
 message FileSystemStoragePathSourceConfig {
+
+  // The policy to defines how many versions of the servable should be
+  // served at the same time
+  enum VersionPolicy {
+    // Only serve the latest version that exists in the base path.
+    LATEST_VERSION = 0;
+    // Serves all the versions that exist in the base path.
+    ALL_VERSION = 1;
```
chrisolston: nit: ALL_VERSIONS (add an S)

tensorflow_serving/sources/storage_path/file_system_storage_path_source.proto
```diff
@@ -12,6 +22,9 @@ message FileSystemStoragePathSourceConfig {
 
     // The path to monitor, i.e. look for child paths of the form base_path/123.
     string base_path = 2;
+
+    VersionPolicy version_policy = 3;
```
chrisolston: Please add a brief comment.

tensorflow_serving/g3doc/serving_advanced.md
```diff
@@ -143,8 +143,11 @@ With all these, `ServerCore` internally does the following:
   the `SessionBundleSourceAdapter`.
 
 Whenever a new version is available, this `AspiredVersionsManager` always
-unloads the old version and replaces it with the new one. If you want to
-start customizing, you are encouraged to understand the components that it
+unloads the old version and replaces it with the new one. The default behavior means that
```
chrisolston: Let's replace the first sentence with:

Whenever a new version is available, this `AspiredVersionsManager` loads the new version,
and, under its default behavior, unloads the old one.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.proto
```diff
@@ -4,6 +4,16 @@ package tensorflow.serving;
 
 // Config proto for FileSystemStoragePathSource.
 message FileSystemStoragePathSourceConfig {
+
+  // The policy to defines how many versions of the servable should be
+  // served at the same time
+  enum VersionPolicy {
+    // Only serve the latest version that exists in the base path.
```
chrisolston: Add: "This is the default behavior."

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -23,6 +23,7 @@ limitations under the License.
 #include "tensorflow/core/lib/io/path.h"
 #include "tensorflow/core/lib/strings/numbers.h"
 #include "tensorflow/core/platform/env.h"
+#include "tensorflow/core/util/env_var.h"
```
chrisolston: Please remove this #include.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,34 +77,50 @@ std::set<string> GetDeletedServables(
...
-  // we don't emit an empty aspired-versions list for a non-existent (or
-  // transiently unavailable) base-path. (On some platforms, GetChildren()
-  // returns an empty list instead of erring if the base path isn't found.)
-  if (!Env::Default()->FileExists(servable.base_path())) {
-    return errors::InvalidArgument("Could not find base path ",
-                                   servable.base_path(), " for servable ",
-                                   servable.servable_name());
-  }
+    std::vector<ServableData<StoragePath>>* versions,
+    const std::vector<string>& children) {
```
chrisolston: Please move 'children' before 'versions' in the argument order. (Sorry, this is a Google style convention.)

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,34 +77,50 @@ std::set<string> GetDeletedServables(
   return deleted_servables;
 }
 
-// Like PollFileSystemForConfig(), but for a single servable.
```
chrisolston: Please document this function: what does it do? what do its arguments mean? (in particular, what does 'children' represent?)

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,34 +77,50 @@ std::set<string> GetDeletedServables(
...
 
-  // Retrieve a list of base-path children from the file system.
-  std::vector<string> children;
-  TF_RETURN_IF_ERROR(
-      Env::Default()->GetChildren(servable.base_path(), &children));
+  LOG(INFO) << "Emit all servable versions";
+
+  // Identify all the versions, among children that can be interpreted as
+  // version numbers.
+  int version_child = -1;
```
chrisolston: I'd rather not assume that version numbers are non-negative.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,34 +77,50 @@ std::set<string> GetDeletedServables(
...
+    if (!strings::safe_strto64(child.c_str(), &child_version_num)) {
+      continue;
+    }
+
+    version_child = i;
+    found_version = child_version_num;
+    // Emit all the aspired-versions data.
+    if (version_child >= 0) {
+      const ServableId servable_id = {servable.servable_name(), found_version};
+      const string full_path =
```
chrisolston: Can we avoid duplicating code between EmitAllVersions() and EmitLatestVersion()?

Maybe have one function that decides which versions to emit (just which 'children' entries), and then a separate one that converts them into ServableData items.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -132,6 +149,44 @@ Status PollFileSystemForServable(
...
+  for (int i = 0; i < children.size(); ++i) {
+    const string& child = children[i];
+    real_children.insert(child.substr(0, child.find_first_of('/')));
+  }
+  children.clear();
+  children.insert(children.begin(), real_children.begin(), real_children.end());
+
+  LOG(INFO) << "Version policy of servable:" << servable.version_policy();
+
+  if (servable.version_policy() == FileSystemStoragePathSourceConfig::ALL_VERSION) {
```
chrisolston: Please use a switch statement. That way if, in the future, we add another case, the compiler will force us to examine this code and consider how to handle the new case.

tensorflow_serving/sources/storage_path/file_system_storage_path_source_test.cc
```diff
@@ -171,6 +171,38 @@ TEST(FileSystemStoragePathSourceTest, MultipleVersions) {
...
+  const string base_path = io::JoinPath(testing::TmpDir(), "MultipleVersionsAtTheSameTime");
+  TF_ASSERT_OK(Env::Default()->CreateDir(base_path));
+  TF_ASSERT_OK(Env::Default()->CreateDir(
+      io::JoinPath(base_path, "non_numerical_child")));
+  TF_ASSERT_OK(Env::Default()->CreateDir(io::JoinPath(base_path, "42")));
+  TF_ASSERT_OK(Env::Default()->CreateDir(io::JoinPath(base_path, "17")));
+
+  auto config = test_util::CreateProto<FileSystemStoragePathSourceConfig>(
+      strings::Printf("servables: { "
+                      "  version_policy: 1 "
```
chrisolston: Please use the enum value (ALL_VERSIONS) rather than 1.

waichee:
@chrisolston that's fine. I am glad that it gets scrutinised, and its helpful to get feedback from main contributor(s) of the repo. I've updated the following as per the comment:
- punctuations and wordings
- add comments and descriptions where required
- make the "EmitAllVersions" and "EmitLatestVersion" as DRY as possible
- use enum in the test

On a separate note, just wondering, for the config protos that users would like to override or customize, what is the best way of doing it? Updating the config values and recompile the source?

chrisolston:
Thanks for making the changes. I'll take another pass.

re: the config -- the best is to use flags. For the ModelServer binary we already have some flags. Right now we don't have flags controlling all aspects of the Source & Manager config, but we're planning to do a clean-up soon that will improve the situation. We'll probably eliminate the current hodge-podge of one-off flags in favor of passing in paths to textual proto config files.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.proto
```diff
@@ -4,6 +4,16 @@ package tensorflow.serving;
 
 // Config proto for FileSystemStoragePathSource.
 message FileSystemStoragePathSourceConfig {
+
+  // The policy to defines how many versions of the servable should be
```
chrisolston: nit: "to define" (no s)

tensorflow_serving/sources/storage_path/file_system_storage_path_source_test.cc
```diff
@@ -171,6 +171,40 @@ TEST(FileSystemStoragePathSourceTest, MultipleVersions) {
...
 
+TEST(FileSystemStoragePathSourceTest, MultipleVersionsAtTheSameTime) {
+  const string base_path = io::JoinPath(testing::TmpDir(), "MultipleVersionsAtTheSameTime");
+  TF_ASSERT_OK(Env::Default()->CreateDir(base_path));
+  TF_ASSERT_OK(Env::Default()->CreateDir(
+      io::JoinPath(base_path, "non_numerical_child")));
+  TF_ASSERT_OK(Env::Default()->CreateDir(io::JoinPath(base_path, "42")));
+  TF_ASSERT_OK(Env::Default()->CreateDir(io::JoinPath(base_path, "17")));
+
+  string all_version_policy = std::to_string(FileSystemStoragePathSourceConfig::ALL_VERSIONS);
```
chrisolston: I don't think you need to do this. You can simply in-line ALL_VERSIONS int he proto text, as in:

servables {
  version_policy: ALL_VERSIONS
  ...

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
   return deleted_servables;
 }
 
-// Like PollFileSystemForConfig(), but for a single servable.
+//Add a new ServableData for the model version to the the vector of versions to aspired
+void UpdateVersionToAspire(
+    const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
+    const string version_relative_path,
```
chrisolston: const string&

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+    const int version_number,
+    std::vector<ServableData<StoragePath>>* versions) {
+    const ServableId servable_id = {servable.servable_name(), version_number};
+    const string full_path =
+        io::JoinPath(servable.base_path(), version_relative_path);
+    versions->emplace_back(ServableData<StoragePath>(servable_id, full_path));
+}
+
+bool IsValidVersionNumber(const string& version_path, int64 *version_number) {
```
chrisolston: Nice!

1. Please rename to ParseVersionNumber.
2. Add documentation along the lines of: attempts to parse ..., upon success writes the result to 'version_number' and returns true; else returns false

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+  return strings::safe_strto64(version_path.c_str(), version_number);
+}
+
+// Update the servable data to include
+// all the model versions found in the base path as aspired versions.
+bool EmitAllVersions(
+    const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
+    const std::vector<string>& children,   // represents a list of base-path children from the file system.
```
chrisolston: Please move the explanation of 'children' to the method comment, as in:

// Update the servable data to include ...
//
// 'children' represents ...
bool EmitAllVersions( ...

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+        io::JoinPath(servable.base_path(), version_relative_path);
+    versions->emplace_back(ServableData<StoragePath>(servable_id, full_path));
+}
+
+bool IsValidVersionNumber(const string& version_path, int64 *version_number) {
+  return strings::safe_strto64(version_path.c_str(), version_number);
+}
+
+// Update the servable data to include
+// all the model versions found in the base path as aspired versions.
```
chrisolston: Explain what the return value represents.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+bool EmitAllVersions(
+    const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
+    const std::vector<string>& children,   // represents a list of base-path children from the file system.
+    std::vector<ServableData<StoragePath>>* versions) {
+
+  LOG(INFO) << "Emit all servable versions";
+
+  bool version_found = false;
+  for (int i = 0; i < children.size(); ++i) {
```
chrisolston: Please drop the 'i', and instead write:
for (const string& child : children) {
  ... child ...

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
   return deleted_servables;
 }
 
-// Like PollFileSystemForConfig(), but for a single servable.
+//Add a new ServableData for the model version to the the vector of versions to aspired
+void UpdateVersionToAspire(
```
chrisolston: Please rename to AspireVersion()

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+}
+
+// Update the servable data to include
+// all the model versions found in the base path as aspired versions.
+bool EmitAllVersions(
+    const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
+    const std::vector<string>& children,   // represents a list of base-path children from the file system.
+    std::vector<ServableData<StoragePath>>* versions) {
+
+  LOG(INFO) << "Emit all servable versions";
```
chrisolston: Please don't add any logging in this PR. We have already cut our logging because it has become too verbose for some users.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+    versions->emplace_back(ServableData<StoragePath>(servable_id, full_path));
+}
+
+bool IsValidVersionNumber(const string& version_path, int64 *version_number) {
+  return strings::safe_strto64(version_path.c_str(), version_number);
+}
+
+// Update the servable data to include
+// all the model versions found in the base path as aspired versions.
+bool EmitAllVersions(
```
chrisolston: Please rename to AspireAllVersions().

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+      version_found = true;
+    }
+  }
+
+  return version_found;
+}
+
+// Update the servable data to include
+// the latest version found in the base path as aspired version.
+bool EmitLatestVersion(
```
chrisolston: AspireLatestVersion().

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+      UpdateVersionToAspire(servable, children[i], version_number, versions);
+      version_found = true;
+    }
+  }
+
+  return version_found;
+}
+
+// Update the servable data to include
+// the latest version found in the base path as aspired version.
```
chrisolston: Document return value.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+    const string full_path =
+        io::JoinPath(servable.base_path(), version_relative_path);
+    versions->emplace_back(ServableData<StoragePath>(servable_id, full_path));
+}
+
+bool IsValidVersionNumber(const string& version_path, int64 *version_number) {
+  return strings::safe_strto64(version_path.c_str(), version_number);
+}
+
+// Update the servable data to include
```
chrisolston: Document return value.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+  int64 latest_version_number;
+  bool version_found = false;
+  for (int i = 0; i < children.size(); ++i) {
+    int64 version_number;
+    if (!IsValidVersionNumber(children[i], &version_number)) {
+      continue;
+    }
+
+    // Check if this is the largest version number
+    if (latest_version_child_index == 0 || latest_version_number < version_number) {
```
chrisolston: Change to:
if (!version_found || ...)

... and drop the =0 in the initialization of latest_version_child_index.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,7 +76,83 @@ std::set<string> GetDeletedServables(
...
+
+  // Emit the latest aspired version.
+  if (version_found) {
+    UpdateVersionToAspire(servable, children[latest_version_child_index], latest_version_number, versions);
+  }
+
+  return version_found;
+}
+
+// Like PollFileSystemForConfig(), but for servable
```
chrisolston: Please revert to original comment:
// Like PollFileSystemForConfig(), but for a single servable.

tensorflow_serving/sources/storage_path/file_system_storage_path_source_test.cc
waichee: ahhh good to know.

waichee:
@chrisolston  Updated the changes as requested. 

Thanks for the info on the config customization. Good to know that there's plan to consolidate the config in the future.  I'll raise a separate PR for a flag on model server to turn on the serving of multiple versions.

waichee:
@chrisolston  would you be able to have a look at this and let me know if any further changes required?

thanks

tensorflow_serving/sources/storage_path/file_system_storage_path_source_test.cc
chrisolston: Please delete "string all_version_policy ..."

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,6 +76,84 @@ std::set<string> GetDeletedServables(
...
+// Attempts to parse the version path as integer, upon success writes the result to 'version_number' and returns true.
+// Otherwise returns false when fail to parse the version path.
+bool ParseVersionNumber(const string& version_path, int64 *version_number) {
+  return strings::safe_strto64(version_path.c_str(), version_number);
+}
+
+// Update the servable data to include
+// all the model versions found in the base path as aspired versions.
+// The argument 'children' represents a list of base-path children from the file system.
+// Returns true if valid servable path is found, otherwise returns false.
```
chrisolston: Please change to "Returns true if one or more valid servable version paths are found, otherwise returns false."

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,6 +76,84 @@ std::set<string> GetDeletedServables(
...
+bool AspireAllVersions(
+    const FileSystemStoragePathSourceConfig::ServableToMonitor& servable,
+    const std::vector<string>& children,   // represents a list of base-path children from the file system.
+    std::vector<ServableData<StoragePath>>* versions) {
+
+  bool version_found = false;
```
chrisolston: Please rename to at_least_one_version_found.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -76,6 +76,84 @@ std::set<string> GetDeletedServables(
...
+  }
+
+  return version_found;
+}
+
+// Update the servable data to include
+// the latest version found in the base path as aspired version.
+// The argument 'children' represents a list of base-path children from the file system.
+// Returns true if valid servable path is found, otherwise returns false.
+bool AspireLatestVersions(
```
chrisolston: Rename to AspireLatestVersion (no 's' at the end).

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -105,30 +183,17 @@ Status PollFileSystemForServable(
...
-    int64 child_version_num;
-    if (!strings::safe_strto64(child.c_str(), &child_version_num)) {
-      continue;
-    }
-
-    if (latest_version_child < 0 || latest_version < child_version_num) {
-      latest_version_child = i;
-      latest_version = child_version_num;
-    }
+  bool version_found = false;
```
chrisolston: Please rename to at_least_one_version_found.

waichee:
Thanks @chrisolston . I've updated those comments, can you have a look again?

chrisolston:
Thanks for making all the changes. I've merged the PR.

waichee:
Thanks Chris for taking the time to review

chrisolston:
Thank *you* for contributing to TF-Serving.

lovejoy:
I have test,  Tensorflow model server doesn't support TF_SERVING_MULTIPLE_VERSIONS at all ?

JGuillaumin:
It does not, you have to use a config file. 
With this example, you will put into serving all versions in `saved_model/`
```
model_config_list: {
  config: {
    name: "CIFAR10",
    base_path: "/mnt/serving_cifar10/saved_model/",
    model_platform: "tensorflow",
    model_version_policy: {
      all {}
    }
  },
}
```

PR #234, opened 2016-11-04
Fix inception and mnist examples for Python 3

jmargeta:
- Ran 2to3 converter to add parentheses for print statement 
 - xrange replaced for range


kirilg:
Jenkins test this please

PR #248, opened 2016-11-15
Fix inception_k8s.yaml filename in examples and docs

jgmize:


vinuraja:
Thanks for the fix @jgmize.

Issue #190, opened 2016-09-27
Servable Traits?


rayglover-ibm:
I noticed that the `tensorflow_model_server` naturally only supports the TensorFlow servable. It seems that the intended approach is for other servables to implement their own model_server, wrapping server_core where sensible.

Some servables are more similar than others, however. I've been working on Caffe support [here](https://github.com/rayglover-ibm/serving-caffe) which mimics the TensorFlow servable implementation. Since there are many similarities, it felt natural to modify the model server to be parameterized on traits of the servable. You can see this attempted [here](https://github.com/rayglover-ibm/serving-caffe/blob/caffe/tensorflow_serving/model_servers/main.cc#L108).

I believe this compositional approach would be beneficial to other parts, in particular the batching componentry, which the last time I looked was quite TensorFlow (and `tensorflow::Tensor`) centric.

Perhaps something else along similar lines has been considered before?

chrisolston:
Yes, as you can see we've tried to be tensorflow-agnostic in most of our code, but at the same time haven't gone overboard with templating to support other kinds of ML models. Extending the model-server to be templetized for the kind of model is a good idea. An alternative approach to templates/traits would be to use a registry pattern. I mention that because inside google we have discussed moving to the registry approach for TF-Serving, although it hasn't bubbled to the top of our priority list yet which is why you haven't seen any movement there :)

I wouldn't want to block your progress so it would be great to move forward with the templates/traits approach. Looking forward to the contribution!

rayglover-ibm:
Admittedly I've noticed the TensorFlow (and perhaps Google generally) avoid the template-all-the-things approach, meaning a registry system is probably more in keeping with the grand scheme of things.

In addition it may be feasible to add some build magic to allow people to plug-in their servable(s). My servable has support for something along these lines by allowing one to specify which fork of Caffe to build with, but I'm thinking more generally.

Anyway I thought I'd let it be known that I'm interested in such things as a means to leverage TensorFlow Serving as a generic model serving framework, and I'd adopt any move in that direction. It would also likely make any attempt to contribute additional servables less painful. :)

Issue #186, opened 2016-09-22
GPU Build Issues Unhandled exception 


promiseofcake:
We are having a few issues building GPU on the latest HEAD. Initially got past some `local_cuda_config` errors by pinning to the latest head last yesterday (https://github.com/tensorflow/tensorflow/commit/a5f8f42). But we are now receiving this error when attempting to build:

Basing on CUDA 7.5, cuDNN 5.1.3.

Bazel 0.3.1

```
# env
CUDA_HOME=/usr/local/cuda
CUDA_PATH=/usr/local/cuda-7.5
CUDA_TOOLKIT_PATH=/usr/local/cuda-7.5
CUDA_VERSION=7.5
LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64:/usr/local/cuda-7.5/targets/x86_64-linux/lib:/usr/lib/x86_64-linux-gnu/
LIBRARY_PATH=/usr/local/cuda/lib64/stubs:
TF_CUDA_COMPUTE_CAPABILITIES=3.0
TF_CUDA_VERSION=7.5
TF_CUDNN_VERSION=5.1.3
TF_NEED_CUDA=1
```

```
bazel build -c opt --config=cuda --verbose_failures tensorflow_serving/...
```

```
INFO: Reading 'startup' options from /root/.bazelrc: --batch
____Loading package: tensorflow_serving/servables/tensorflow/testdata
____Loading...
Unhandled exception thrown during build; message: Unrecoverable error while evaluating node 'CONFIGURATION_FRAGMENT:com.google.devtools.build.lib.skyframe.ConfigurationFragmentValue$ConfigurationFragmentKey@8db88dd6' (requested by nodes 'CONFIGURATION_COLLECTION:com.google.devtools.build.lib.skyframe.ConfigurationCollectionValue$ConfigurationCollectionKey@54fe1e0', 'CONFIGURATION_FRAGMENT:com.google.devtools.build.lib.skyframe.ConfigurationFragmentValue$ConfigurationFragmentKey@e90862b7', 'CONFIGURATION_FRAGMENT:com.google.devtools.build.lib.skyframe.ConfigurationFragmentValue$ConfigurationFragmentKey@ff24d609')
____Elapsed time: 1.529s
[...]

PR #261, opened 2016-11-29
Caffe support

rayglover-ibm:
This contains just the core implementation of the fork I've been maintaining [here](https://github.com/rayglover-ibm/serving-caffe) under `tensorflow_serving/servables/caffe/...` plus some refactoring of things under `tensorflow_serving/model_servers/...`, which I mentioned in #190 .

The idea from my perspective is to make it easier for teams using [Caffe](https://github.com/BVLC/caffe) in production to move to a more ML-platform agnostic serving framework; that seems to fit nicely with the aims of TensorFlow Serving!

__Summary:__

- The Caffe CMake build is driven by Bazel
- Supports CUDA by adopting the TensorFlow CUDA configuration
- Supports Python layers by python interpreter integration
- The implementation tries to follow the TensorFlow integration; I think this has been a success even though other parts of Tensorflow Serving are fairly `Tensor` specific (e.g. the batching infrastructure)
- Switching between forks of Caffe when building can be useful if you have multiple services each requiring different custom layers or optimizations within Caffe itself; there is a macro `caffe_pkg` in the build to make this easier.

__Additional things I've included:__

- A Dockerfile for creating Caffe dev-env (`tensorflow_serving/tools/docker/Dockerfile.caffe-devel`)
- A fix to the CUDA crosstool I mentioned [here](https://github.com/tensorflow/serving/issues/186#issuecomment-251152755). Not sure what the status of this problem is.  

__TODO / Caveats:__

- As part of the mnist tutorial a pre-trained model is downloaded from `ibm.box.com`, it's probably worthwhile moving this elsewhere. Alternatively the model could be trained on the local machine but this would require a more thought-out approach.

- I think the Python integration is good enough but hardly watertight; I can imagine issues arising here if one tries to serve more than one caffe net (with Python layers in each) simultaneously. Using sub-interpreters (one per net) should make this more robust.

- The build isn't guaranteed to adopt the Python libs/includes set in the tensorflow configuration; instead cmake will attempt to find python itself. 

- There isn't any user documentation in this PR -- I'm looking for advice on this.

__Things I've removed (but you might be interested to know about):__

- There is an example object detector [here](https://github.com/rayglover-ibm/serving-caffe#example-object-detection) supporting _[Faster R-CNN](https://github.com/rbgirshick/py-faster-rcnn)_ and _[SSD: Single Multibox Detector](https://github.com/weiliu89/caffe/tree/ssd)_ object detection architectures; this demonstrates use of multiple Caffe forks, Python interpreter integration, image transpositions and means-subtraction on Tensor types.


perone:
Are there any plans to merge this pull request ?

rayglover-ibm:
I also wouldn't mind being abreast of any plans to consider this PR; at least then I'd know whether to put more time in to maintaining it. Perhaps @chrisolston knows?

kirilg:
Thanks for the PR and sorry for the delay!

It's really great to see such a change to integrate Caffe with TensorFlow Serving.

We discussed it with the team, and had some concerns around maintaining the code that will now depend on a new external submodule. I think the best course of action is to fork the repo as you've done, and keep all caffe specific code there. We'd still love to accept all caffe independent changes into this repo, which will hopefully make it easier for you to maintain your fork long term.

To make it easier, please take a look at the class_registration util that was added in [this commit](https://github.com/tensorflow/serving/commit/f83efeee8d20b705826103f8d69ef4cc996ad4e8). We now have a registry for source adapters, which lets you have custom code that creates a servable. The base class will still have to be SavedModelBundle, but I think your implementation of CaffeSessionBundle can adhere to that.

It would also be great to get your cuda fix in a separate PR. Thanks again!

Issue #288, opened 2017-01-10
Raise the max message size of gRPC for tensorflow model server


tobegit3hub:
We have used TensorFlow Serving to load multiple models for online or offline services. For online services, we focus on the latency of each gRPC request and it's good enough. But for offline services, we want to improve the throughout by batching data in one gRPC request.

Then we need the limitation from gRPC max message size which is already discussed in https://github.com/tensorflow/serving/issues/284 . The default max size is 4M which is much smaller than the real world requests.

I'm not sure if 100M is suitable but it would be better for most users. I may send an PR  to raise the max message size if someone also think it's reasonable. Otherwise, most developers need to compile the source code to change this limitation.

kirilg:
Discussed on the [PR](https://github.com/tensorflow/serving/pull/289), but our current limit as of [this commit](https://github.com/tensorflow/serving/commit/543c89517cdcf944e79b3cee97aed0fcb1679a16) is 2G (as of about a month ago). Perhaps you're running with an old binary that's still using the default 4M? We tested the model_server after the change and it should be able to handle requests of 100M already.

tobegit3hub:
It may be my mistake and thanks @kirilg for the detailed explanation. 

I have compiled the latest tensorflow_ model_server and the restriction should be in clients.

PR #289, opened 2017-01-10
Raise gRPC max message size from 4M to 100M

tobegit3hub:
Resolve https://github.com/tensorflow/serving/issues/288

### DIFF
--- a/tensorflow_serving/model_servers/main.cc
+++ b/tensorflow_serving/model_servers/main.cc
@@ -200,7 +200,8 @@ void RunServer(int port, std::unique_ptr<ServerCore> core,
   std::shared_ptr<grpc::ServerCredentials> creds = InsecureServerCredentials();
   builder.AddListeningPort(server_address, creds);
   builder.RegisterService(&service);
-  builder.SetMaxMessageSize(tensorflow::kint32max);
+  // Raise max frame size for larger request if it is reasonable
+  builder.SetMaxMessageSize(104857600);
   std::unique_ptr<Server> server(builder.BuildAndStart());
   LOG(INFO) << "Running ModelServer at " << server_address << " ...";
   server->Wait();


tensorflow_serving/model_servers/main.cc
```diff
@@ -200,7 +200,8 @@ void RunServer(int port, std::unique_ptr<ServerCore> core,
   std::shared_ptr<grpc::ServerCredentials> creds = InsecureServerCredentials();
   builder.AddListeningPort(server_address, creds);
   builder.RegisterService(&service);
-  builder.SetMaxMessageSize(tensorflow::kint32max);
+  // Raise max frame size for larger request if it is reasonable
+  builder.SetMaxMessageSize(104857600);
```
kirilg: I'm confused, the original wasn't 4M, it's 2G isn't it?

1Feng: static const int32 kint32max = ((int32)0x7FFFFFFF);
It's 4M.
2GB is the ProtoBuf's hard limit.

supertim:
magic number

tensorflow_serving/model_servers/main.cc
kirilg: kint32max == ((int32)0x7FFFFFFF) == 2G, not 4M. What am I missing?

1Feng: I am sorry. It's my  fault. 2G is  right.

kirilg: Just to confirm, are you running with TensorFlow Serving at head and are running into a problem that the requests are too large? I think this PR should not be necessary because the current limit is greater than the newly proposed 100M. When we added this change about a month ago in [this commit](https://github.com/tensorflow/serving/commit/543c89517cdcf944e79b3cee97aed0fcb1679a16), we tested with requests that were in the order of 100M and it worked (and of course failed with the old default of 4M).

PR #290, opened 2017-01-10
Add python predict client

tobegit3hub:
Add the instructions about how to compile proto files and the independent inception example.

### DIFF
--- a/python_predict_client/README.md
+++ b/python_predict_client/README.md
@@ -1,0 +1,29 @@
+# Python Predict Client
+
+It is the independent Python predict client for TensorFlow Serving.
+
+## Installation
+
+Copy the proto files.
+
+```
+cp ../tensorflow_serving/apis/{model,predict,prediction_service}.proto .
+```
+
+Replce the import path.
+
+```
+sed -i "s/tensorflow_serving\/apis\///g" {predict,prediction_service}.proto
+```
+
+Comppile the proto files.
+
+```
+python -m grpc.tools.protoc --proto_path=../tensorflow --proto_path=. --python_out=. --grpc_python_out=. {model,predict,prediction_service}.proto
+```
+
+## Inception Client
+
+```
+python ./inception_client.py --server=127.0.0.1:9000 --image=foo.jpg
+```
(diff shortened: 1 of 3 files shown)


tobegit3hub:
For most python developers, they don't `bazel` to manage dependencies. Developers can compile the proto files by ourselves with `protoc` or `grpcio`. We have used this library in production for our Python programs and this batteries-included independent client could be really helpful for other users.

tobegit3hub:
I would not like to move and change the import path of the proto files, but the generated Python libraries are not complete. 

I think it's the easiest way for user to generate proto files and import to use in this way. Any feedback is welcome.

python_predict_client/README.md
```diff
@@ -0,0 +1,29 @@
...
+Copy the proto files.
+
+```
+cp ../tensorflow_serving/apis/{model,predict,prediction_service}.proto .
+```
+
+Replce the import path.
+
+```
+sed -i "s/tensorflow_serving\/apis\///g" {predict,prediction_service}.proto
```
kirilg: Thanks for the contribution!

Having to `cp` and `sed` source files just to get it to work seems hard to maintain and might be problematic in some cases. This client still uses Tensorflow, so it appears the main benefit of this is being able to call `python ...` without going through Bazel. We already rely on Bazel very heavily in this repo and I think for these example clients, it's not too unreasonable to use bazel-bin/ even for python (as the original clients do), considering that you already need Bazel to build the model_server itself.

I think the best course of action is to comment on the GitHub issue, possibly linking to your external repo that contains an example so others who want to avoid Bazel can use it. In this repo though, we'd like to keep the original client which uses Bazel that's easier for us to test and maintain. We recommend this as the officially supported and tested approach.

tobegit3hub: Thanks for the comment. It's reasonable and we don't want to add more dependencies than Bazel. The native python client for TensorFlow Serving is always needed and we have this project to manage it in https://github.com/tobegit3hub/deep_recommend_system/tree/master/python_predict_client .

I'm going to close this pull-reqeust and re-open if we have the better way to do that 😃

Issue #45, opened 2016-04-16
Deploying multiple models


viksit:
What would be the best way to load and serve multiple models into the system?

Looking at the code that exists today, I couldn't see anything directly usable. If there's nothing that does this today, what would be a good set of things to look at to implement something like the following,

Some thoughts,
- Lets assume each model resided in a directory called /mnt/modelid/0000001
- The current `FileSystemStoragePathSource` only looks under a given path for the aspired version. So ideally, there are some methods somewhere that can instead help look for modelid rather than within it.
- The output of this should be a hashmap of `servableId` type objects, and I ought to be able to do a `Get()` from such an object to get the right `modelId`, and then run `classify()` on it.

Lastly - do you guys have use cases where there are multiple models spread out across multiple machines, and using some sort of a DHT implementation to access them?

chrisolston:
Hi there,

The simplest approach is to deploy N instances of FileSystemStoragePathSource. The upstream modules that "catch" the aspired-versions calls are thread-safe (although I don't think we have good testing around that -- perhaps something you could contribute?).

If N gets large that starts to get ridiculous. Also, if you don't know the set of model names (a.k.a. servable names) in advance it wouldn't work because each FileSystemStoragePathSource requires to be configured with a model name. It would be great to have an alternative or generalized file-system source that would look for model_name/version_number, as you say. A contribution on that front would be welcome!

Regarding models on multiple machines -- we don't have anything that explicitly helps with that scenario at the moment. Perhaps if you can elaborate on your requirements and/or proposed architecture we can discuss further. It might be worth moving that to a different thread -- it's a distinct topic. Thanks.

Chris

viksit:
Thanks for the answer, @chrisolston. 

What I've ended up doing is create my own version of a `FileSystemStoragePathSource` that allows me to discover models within a filesystem, which are keyed by ID. I've also made a new version of the `AspiredVersionsManager` which allows me to load/unload these according to an eager policy. This gives me flexibility of either having a model with the same graph loaded with different datasets (versions), or different model graphs + datasets keyed by IDs.

Will start another thread around the multiple model serving, thanks!

chrisolston:
Cool. Please consider contributing your code back to the project :). We are open to pull requests.

viksit:
Absolutely :) - let me put these through the paces first. 

One issue that would be good to get more info on is https://github.com/tensorflow/serving/issues/46.

[...]

PR #294, opened 2017-01-13
Deploying multiple models

phvu:
As discussed in https://github.com/tensorflow/serving/issues/45, add a `model_config_file` argument to tensorflow_serving, which is a file to a `ModelServerConfig` protobuf.


perdasilva:
I'm okay with the commits being contributed to this project ^^

tensorflow_serving/model_servers/main.cc
```diff
@@ -128,6 +131,20 @@ ModelServerConfig BuildSingleModelConfig(
   return config;
 }
 
+ModelServerConfig BuildModelConfigFromFile(
+    const string& config_file_path) {
+
```
chrisolston: Can you please follow the pattern used for ParsePlatformConfigMap() (in particular, using the tensorflow Env to read the file rather than open()). Perhaps you can create a utility function that both BuildModelConfigFromFile() and ParsePlatformConfigMap() can call into. Thanks.

(And accordingly please eliminate #include <fcntl.h>.)

tensorflow_serving/model_servers/main.cc
```diff
@@ -203,26 +220,33 @@ int main(int argc, char** argv) {
...
+                       "If non-empty, read an ascii ModelServerConfig "
+                       "protobuf from the supplied file name, and serve the "
+                       "models in that file. (If used, --model_name, "
+                       "--model_base_path and --model_version_policy "
+                       "are ignored.)"),
       tensorflow::Flag("model_name", &model_name, "name of model"),
+      tensorflow::Flag("model_base_path", &model_base_path,
+                       "path to export (required)"),
       tensorflow::Flag(
-          "model_version_policy", &model_version_policy,
```
chrisolston: It would be great if you could eliminate this formatting change, because it makes it hard for us to merge on the google side. Thanks!

tensorflow_serving/model_servers/main.cc
```diff
@@ -203,26 +220,33 @@ int main(int argc, char** argv) {
...
   tensorflow::string model_version_policy =
       FileSystemStoragePathSourceConfig_VersionPolicy_Name(
           FileSystemStoragePathSourceConfig::LATEST_VERSION);
   std::vector<tensorflow::Flag> flag_list = {
       tensorflow::Flag("port", &port, "port to listen on"),
       tensorflow::Flag("enable_batching", &enable_batching, "enable batching"),
+      tensorflow::Flag("model_config_file", &model_config_file,
+                       "If non-empty, read an ascii ModelServerConfig "
+                       "protobuf from the supplied file name, and serve the "
+                       "models in that file. (If used, --model_name, "
```
chrisolston: Can you add comments to each of {model_name, model_base_path, model_version_policy} stating that they are ignored if model_config_file is supplied.

perdasilva: I hope you meant in code (i.e. // comments) rather than in the help text...=S

perdasilva:
I've tried to address all of the comments. Please re-review =)

tensorflow_serving/model_servers/main.cc
chrisolston: No, I meant in the help text.

chrisolston: Please add to the quoted text. Thanks.

tensorflow_serving/model_servers/main.cc
```diff
@@ -101,6 +101,17 @@ using tensorflow::serving::PredictionService;
 
 namespace {
 
+void ParseProtoTextFile(const string& file, google::protobuf::Message* message) {
```
chrisolston: nit: how about having this function return Status, and the callers can use QCHECK if they want.

tensorflow_serving/model_servers/main.cc
```diff
@@ -128,6 +139,18 @@ ModelServerConfig BuildSingleModelConfig(
   return config;
 }
 
+ModelServerConfig BuildModelConfigFromFile(
+    const string& file) {
+
+  ModelServerConfig config;
```
chrisolston: Delete this line?

tensorflow_serving/model_servers/main.cc
perdasilva: Do you mean tensorflow::Status?
It seems it doesn't work with QCHECK. Would then TF_CHECK need to be used?

What I've done is change it to return a bool. This way it can be used with QCHECK.
Sorry, I'm new to the code base (and C++ =S)

Should you still prefer tensorflow::Status, would the correct implementation be to check if ParseFromString returns true and return Status::OK() or errors::Unknown(...) or some other error. If so, which error?

Thanks,

Per

perdasilva: I've pushed with the bool version. If you still want Status just let me know (with a little guidance) and I'll fix it up.

Thank you!

tensorflow_serving/model_servers/main.cc
```diff
@@ -101,6 +101,16 @@ using tensorflow::serving::PredictionService;
 
 namespace {
 
+bool ParseProtoTextFile(const string& file, google::protobuf::Message* message) {
```
chrisolston: Please return tensorflow::Status. Callers can use TF_CHECK_OK. Thanks.

tensorflow_serving/model_servers/main.cc
```diff
@@ -101,6 +101,16 @@ using tensorflow::serving::PredictionService;
...
+  std::unique_ptr<tensorflow::ReadOnlyMemoryRegion> file_data;
+  TF_CHECK_OK(  // Crash ok
```
chrisolston: TF_RETURN_IF_ERROR

tensorflow_serving/model_servers/main.cc
```diff
@@ -128,6 +138,17 @@ ModelServerConfig BuildSingleModelConfig(
   return config;
 }
 
+ModelServerConfig BuildModelConfigFromFile(
+    const string& file) {
+  ModelServerConfig config;
```
chrisolston: Delete this line? You don't use 'config' anywhere.

tensorflow_serving/model_servers/main.cc
```diff
@@ -101,6 +101,16 @@ using tensorflow::serving::PredictionService;
 
 namespace {
```
chrisolston: Can you please add test coverage of multiple models to tensorflow_model_server_test.py.

perdasilva:
I've tried to address all the comments. Not sure my approach for testing a bad configuration file is the best...let me know if you have a better idea...

I also had to use a placeholder for the test directory, in order to get the configuration file. Hopefully that's ok...

tensorflow_serving/model_servers/BUILD
```diff
@@ -141,6 +141,11 @@ py_test(
         "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two/00000123/export.data-00000-of-00001",
         "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two/00000123/export.index",
         "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two/00000123/export.meta",
+        "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two2/00000123/export.data-00000-of-00001",
```
chrisolston: How about making it half_plus_three, and making it output 0.5*x + 3 ... so we can be sure the model server is running the right model (half-plus-two versus half-plus-three).

tensorflow_serving/model_servers/BUILD
```diff
@@ -141,6 +141,11 @@ py_test(
...
+        "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two2/00000123/export.index",
+        "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two2/00000123/export.meta",
+        "//tensorflow_serving/servables/tensorflow/testdata:good_config.template",
```
chrisolston: How about .txt as the extension for both good_config and bad_config.

Also, please make it good_model_config / bad_model_config because there are multiple kinds of configs.

tensorflow_serving/model_servers/tensorflow_model_server_test.py
```diff
@@ -163,7 +205,41 @@ def _TestBadModelUpconvertedSavedModel(self):
   def _TestBadModelSessionBundle(self):
     """Test Predict against a bad SessionBundle model export."""
     self._TestBadModel(use_saved_model=False)
-
-
+    
+  def testGoodFileConfig(self):
```
chrisolston: Can you name it "testGoodModelConfig" (since there are multiple kinds of configs). Thanks.

tensorflow_serving/model_servers/tensorflow_model_server_test.py
```diff
@@ -163,7 +205,41 @@ def _TestBadModelUpconvertedSavedModel(self):
...
+    """Test server model configuration from file works with valid configuration"""
+    atexit.register(self.TerminateProcs)
+    model_server_address = self.RunServerWithModelConfigFile(PickUnusedPort(),
+                                                             self._GetGoodConfigFile(),
+                                                             False) # use_saved_model
```
chrisolston: Can you set use_saved_model=True. We are preferring SavedModel in new code and will soon deprecate SessionBundle.

perdasilva:
Done. I've added the half_plus_three model to the tf_serving side (the saved model half_plus_two is on the tf side). I've called it saved_model_half_plus_three to make it clear. Hopefully that's ok.

question (outside this PR): I've make a simple gRPC configuration service and attached it to TFS so that you can remotely reconfigure the service (basically just takes ModelServerConfig as the message and calls ServerCore.ReloadConfig) - if this a useful enough feature for the project? I can polish it and create a PR for that as well...

chrisolston:
I'll take a look at your edits, thanks.

Regarding an RPC to reconfigure the server -- that does indeed sound
useful. Looking forward to the PR for that! Thanks.

tensorflow_serving/model_servers/BUILD
```diff
@@ -141,6 +141,15 @@ py_test(
         "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two/00000123/export.data-00000-of-00001",
         "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two/00000123/export.index",
         "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two/00000123/export.meta",
+        "//tensorflow_serving/servables/tensorflow/testdata:half_plus_two2/00000123/export.data-00000-of-00001",
```
chrisolston: Do you still need half_plus_two2? I was hoping we could just have:
 1. The original half_plus_two
 2. The new half_plus_three
(and nothing else)

tensorflow_serving/model_servers/tensorflow_model_server_test.py
```diff
@@ -81,14 +97,29 @@ def RunServer(self, port, model_name, model_path, use_saved_model):
...
+    
+    print command
+    self.server_proc = subprocess.Popen(shlex.split(command), stderr=pipe)
+    print 'Server started'
+    return 'localhost:' + str(port)
+
   def VerifyPredictRequest(self,
                            model_server_address,
+                           model_name='default',
+                           y_value=3.0,
```
chrisolston: Please omit the default value (3.0) for y_value. The tests would be more readable if it's specified explicitly at each callsite.

chrisolston: Also, nit: can we call it "expected_output" instead of y_value please?

tensorflow_serving/model_servers/BUILD
perdasilva: That was an oversight on my part. I had originally planned on doing both SessionBundle and a SavedModel tests - but then I realised later that the saved model version was in the tensorflow code base...=P good eye!

perdasilva: My bad. Oversight =S

perdasilva:
Done. I'll see about the configuration service if I survive this round haha!

chrisolston:

chrisolston:
Done with the review. Looks like you have to merge some conflicts. Also you'll need to sign the CLA.

perdasilva:
OK - Conflicts resolved - hopefully.

It says above "It's up to you to confirm consent of the commit author(s) and merge this pull request when appropriate" - I'm not sure how you can confirm - but I've written a comment to that effect...if that sufficient?

calhamdower:
@perdasilva I'd love to see your gRPC client, as I seem to be doing the exact same thing as you. C++ is a bit rusty (isn't everyones) but happy to help also.

perdasilva:
@calhamdower I didn't quite get the question. I assumed it was regarding the gRPC service for configuration. If not, ignore the rest of this post and clarify a little more what you mean =)

my c++-fu is also still at grasshopper level, but exposing a small service is not too difficult.

To create the configuration service I basically created a service proto (in the apis dir):

```proto
syntax = "proto3";

package tensorflow.serving;
option cc_enable_arenas = true;

import "tensorflow_serving/config/model_server_config.proto";
import "tensorflow_serving/apis/configuration_response.proto";

// open source marker; do not remove
// ConfigureService provides the facility to configure the server
// at runtime
service ConfigurationService {
  // Configure -- Configures a Service.
  rpc Configure(ModelServerConfig) returns (ConfigurationResponse);

}
```
 and response proto:

```proto
syntax = "proto3";

package tensorflow.serving;
option cc_enable_arenas = true;

// Metadata for an inference request such as the model name and version.
message ConfigurationResponse {

  // Was the configuration successful
  bool success = 1;
  
  // The error message if unsuccessful  
  string error_message = 2;
}
```
NOTE: The return message here is a little stupid. Since if the call fails you get an error in the gRPC client with the message. But I only realises this after and haven't refactored the code...so for now, if everything goes well, you just get a success: true message back...

I also updated the BUILD file to include these files in the apis directory as well as in the model_server directory.

then in main.cc I created the implementation for the service:

```c++
class ConfigurationServiceImpl final : public ConfigurationService::Service {
  public:
	explicit ConfigurationServiceImpl(std::unique_ptr<ServerCore>* core)
	  : core_(core) {}

	grpc::Status Configure(ServerContext* context, const ModelServerConfig* request,
			ConfigurationResponse* response) override {
		LOG(INFO) << "Attempting to reconfigure..." << std::endl;
		if(core_->get()) { // check server core is there
			const grpc::Status status = ToGRPCStatus(core_->get()->ReloadConfig(*request));

			if (!status.ok()) {
				VLOG(1) << "Configuration Failed: " << status.error_message();
			}

			response->set_success(status.ok());
			return status;
		}

		return grpc::Status::OK; // should probably return fail here...
	}

  private:
	std::unique_ptr<ServerCore>* core_;
};
```

and attached it to the server:

```c++
void RunServer(int port, std::unique_ptr<ServerCore> core,
               bool use_saved_model) {
  ...
  ConfigurationServiceImpl configuration_service(&core);;
  ...
  builder.RegisterService(&prediction_service);
  builder.RegisterService(&configuration_service);
 ...
}
```

Note that I had to pass a pointer to the unique_ptr<ServerCode> (instead of the unique_ptr) to the creation of the services. I found that the pointer went to null after passing it to the PredictionServiceImpl... I'm not sure that this is the correct way to deal with this.

For the client, the only way I managed to get it to work was to export all of the tensorflow and tensorflow_serving protos (along with their directory structure) to another folder and compile everything to python libraries. I might have also had to add \_\_init\_\_.py files in each directory so that they would get picked up as modules.

If you need a c++ client for this I'm pretty sure you can follow the same pattern but instead of spitting out python you spit out c++.

But I'm fairly sure it should be possible to do it via bazel. But I haven't had the time to play around with it. Bazel is also fairly new to me.

Then I created a very simple command line tool for testing it:

```python
import argparse
from tensorflow_serving.config import model_server_config_pb2 as model_server_config_pb2
import tensorflow_serving.apis.configuration_service_pb2_grpc as configuration_service_pb2
import sys
from google.protobuf.text_format import Merge
import grpc

parser = argparse.ArgumentParser(description="TFS Configurator")
parser.add_argument('--server', help="host:port", required=True)
parser.add_argument('--config', help="config file", required=True)


def get_configuration_from_file(config_file):

  with open(config_file, 'r') as f:
    data = str(f.read())
    model_server_config = model_server_config_pb2.ModelServerConfig()
    Merge(data, model_server_config)

  f.close()
  return model_server_config

if __name__ == '__main__':
  args = parser.parse_args()

  channel = grpc.insecure_channel(args.server)
  stub = configuration_service_pb2.ConfigurationServiceStub(channel)

  configuration = get_configuration_from_file(args.config)

  print('Server: %s' % args.server)
  print('Configuration: ')
  print(configuration)
  print('Result: ')
  print(stub.Configure(configuration))
```

The code is far from perfect. I only had a spare couple of hours to play around and most of it was spent on trying to get the python client going and messing with BUILD files =P

I hope this helps ^^

calhamdower:
That would have taken me days not hours. 
I actually was implementing something very similar but hadn't yet finished. I think Ill take your work above and run with that! 
Thanks for the detailed response @perdasilva. Im sure this will help other people too; as having a static modelserver definition (as far as served models) at runtime mightn't work for a lot of use-cases. 
Time to battle Bazel and see if I can get this compiled.

perdasilva:
@calhamdower no problems, mate. Glad to help.
I was just thinking about how to build the client and it occurred to me that there should already be an example in the code base. Just look how bazel builds the inception or mnist client in the examples folder. That should provide some hints ^^ If I have some time soon I'll build out the configuration service and submit a PR. All the best!

PR #309, opened 2017-02-03
added --use_saved_model=false

### LINKED ISSUE #308: Tensorflow Serving basic example fails without --use_saved_model=false
I'm unable to run the 'Test the Server' code 
```
$>bazel-bin/tensorflow_serving/example/mnist_client --num_tests=1000 --server=localhost:9000
```
shown in the [basic example ](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/g3doc/serving_basic.md) without producing the error: 

```
AbortionError(code=StatusCode.NOT_FOUND, details="FeedInputs: unable to find feed output images")
```

I've provided the steps that I've followed [here](http://stackoverflow.com/questions/41721526/tensorflow-serving-basic-example).

I can successfully test the server if I including the option `--use_saved_model=false` when starting the server, but I'm not certain this is the correct solution.

ras44:
resolves #308

### DIFF
--- a/tensorflow_serving/g3doc/serving_basic.md
+++ b/tensorflow_serving/g3doc/serving_basic.md
@@ -144,7 +144,7 @@ With that, your TensorFlow model is exported and ready to be loaded!
 
 ~~~shell
 $>bazel build //tensorflow_serving/model_servers:tensorflow_model_server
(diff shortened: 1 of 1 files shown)


ras44:
I signed it!

lilao:
I don't think this change will work with mnist_saved_model.py. Please see my comments in #308.

PR #321, opened 2017-02-13
Updated tensorflow submodule

kchodorow:
I'm working on fixing bazelbuild/bazel#1262, which will change the way Bazel organizes the execution root. Tensorflow had to be updated to work with this change (see https://github.com/tensorflow/tensorflow/pull/7165), so this updates the serving project to use a compatible version of tensorflow.


kirilg:
Jenkins, test this please.

PR #339, opened 2017-03-01
Adds public visibility to predict_impl cc_library

perdasilva:
I'm trying to use serving as a submodule. When building it complained that //tensorflow_serving/servables/tensorflow:predict_impl was not visible. I've added the public visibility to it.

I hope this is ok.


perdasilva:
@chrisolston any chance you could look at this? It's a 3 line change and it would _really_ help me out =)

kirilg:
Jenkins, test this please.

PR #364, opened 2017-03-21
Update README.md

yaolu:
Fix the format error of Markdown in readme file.


yaolu:
The wrong format of README.md file result in the following view.
screenshot
screenshot

So I fix the error of README.md file

chrisolston:
Thanks for the fix!

Issue #354, opened 2017-03-13
Can't import inception_model


mkemka:
This is a basic question and the issue may be my inexperience with docker.
I have followed the tutorial directly on this page:

When I run this step:
bazel-bin/tensorflow_serving/example/inception_saved_model --checkpoint_dir=inception-v3 --output_dir=inception-export

I get a basic import error:
   from inception import inception_model
ImportError: cannot import name inception_model

I can see the inception_model file in the tf_models subdirectory but I can't seem to add it to the path for the function to pick up.

abdasgupta:
Exactly the same problem I am facing too

mkemka:
I thought the fix would be to copy the models to the PYTHONPATH but it still does not seem to recognise it.

NicolasAlmeida:
I face the same problem. Do you already know a possible fix in order to not get this error?

ybbaigo:
+1

ptrblck:
Hi,

I guess it is not the best way to get around this issue, but I tried the following:

* Pull the tensorflow models 
`git pull http://github.com/tensorflow/models`

The `ImportError` is thrown from
 /serving/bazel-bin/tensorflow_serving/example/inception_saved.model.runfiles/tf_serving/tensorflow_serving/example/inception_saved_model.py, 
which is a symbolic link to /serving/tensorflow_serving/example/inception_saved_model.py.

* Create symbolic links from the missing inception model files to /serving/tensorflow_serving/example:
`ls -s ~/models/inception/inception/ /serving/tensorflow_serving/example/inception`

Now the files should be in the right location, but python cannot import them as a module, because the `__init__.py` files are missing in the directories.

* Create `__init__.py` files, to let python import the desired files:
`cd /serving/tensorflow_serving/example/inception
touch __init__.py
cd slim
touch __init__.py`

After these steps, I could succesfully run
`bazel-bin/tensorflow_serving/example/inception_saved_model --checkpoint_dir=inception-v3 --output_dir=inception-export`

I hope it helps you guys!

azagovora:
Thank you!!!

mohan-chinnappan-n:
Thank you, @ptrblck . Your steps fixes this issue! 👍 

Can we fix this in docker image itself using the steps: @ptrblck: 

```
1. cd ; git clone https://github.com/tensorflow/models
2. ln -s ~/models/inception/inception/ /serving/tensorflow_serving/example/inception

[...]

PR #366, opened 2017-03-21
Fix issue with Inception saved model - redo pull request - issue 354

mountaintom:
When the inception repository name is the same as the inception model workspace name this results in anomalous behavior of tensorflow_serving/example build. 
Running the command 
```
bazel-bin/tensorflow_serving/example/inception_saved_model --checkpoint_dir=inception-v3 --output_dir=inception-export
```
fails because a needed support module cannot be loaded.

When the repository name is different than the workspace, a build time warning is generated.

This pull request provides a way to have the repository name be different than the inception workspace name (allowing the examples' build to work) and correct for the build time workspace/repository name mismatch warning.

This is similar to the solution suggested in [https://github.com/bazelbuild/bazel/issues/1331](https://github.com/bazelbuild/bazel/issues/1331)

 Tensorflow-serving issue 354: [https://github.com/tensorflow/serving/issues/354](https://github.com/tensorflow/serving/issues/354)

-- Redoing pull request after fixing author email address in my original commits. --

### DIFF
--- a/tensorflow_serving/example/BUILD
+++ b/tensorflow_serving/example/BUILD
@@ -80,7 +80,7 @@ py_binary(
         "inception_export.py",
     ],
     deps = [
-        "@inception//inception",
+        "@inception_model//inception",
         "@org_tensorflow//tensorflow:tensorflow_py",
         "@org_tensorflow//tensorflow/contrib/session_bundle:exporter",
     ],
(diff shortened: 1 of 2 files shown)


mountaintom:
Even after doing a "git commit --amend --author " and re-pushing to my gethub repository. The GitHub push is not going to let the user@localhost email to be forgotten. So, looks like the bad "cla" is not fixable on my end.

PR #382, opened 2017-03-29
Fix gcloud docker push command

christinakayastha:
Based on documentation at: https://cloud.google.com/container-registry/docs/pushing


kirilg:
I merged your other PR to master, thanks for the fix! We push to gh-pages directly from master, so I'll close this one unmerged.

PR #404, opened 2017-04-14
Update Dockerfile.devel

paristj:
.txt Curls a 404


paristj:
I signed it!

chrisolston:
Thank you for this fix. It turns out this exact change has already been applied at head, so I'll close this PR. (But please do keep these coming in the future!)

PR #419, opened 2017-04-24
Add model_signature_name command line parameter to inception cpp client

mountaintom:
When run, the example cpp inception client fails with the error: 
gRPC call return code: 3: input tensor alias not found in signature: images

The signature name “predict_images" must be explicitly set when the server request is made.

Added code to set the signature name from command line. Will default to "predict_images” if not specified on command line.


lilao:

chrisolston:
Hi,

Sorry for the delay; we had some hiccups on your side this past week. Can you please resolve the conflicts in the wake of https://github.com/tensorflow/serving/commit/da6a36cc55d8bfd77ddd37f13486401ea71a2fd4

Then we'll re-verify everything and merge.

Thanks!

Chris

mountaintom:
Here is an update with the merge conflicts resolved.
Welcome.

PR #442, opened 2017-05-15
UPD inception_k8s.yaml with coherent metadata / selector

bathizte:
Deployment and Service definition use different keys, so that service exposition failed (No pods associated with inception-service).


bathizte:
Signed the CLA

xuchenCN:
Hi @bathizte @kirilg  I working on PR #468 the similar to this PR , I test the yaml file that you committed but I got unexpected result the **Endpoints** of the **Service** is **<none>**

PR #447, opened 2017-05-18
Create TensorFlow Serving clients with minimal build requirements

mountaintom:
Project to build clients with just the resources necessary for clients. 

Also, a jumping off point for developers who would like to port TensorFlow Serving clients to platforms like Android or create FFI style interfaces to other languages.


chrisolston:
Hi there,

Sorry for the slow response. We on the TF-Serving team have been a bit torn about how to handle cases like this. We'd very much like to stay Bazel-only because it greatly reduces the maintenance requirement. But we recognize that Bazel is not everybody's preferred approach. Down the road, if resources permit, we may consider a Makefile alternative if it handles the server-side, so folks don't need to contend with *both* Bazel and Makefile.

-Chris

mountaintom:
Thanks for your consideration of this pull request.

I do want to say that the primary theme of this contribution in not a Makefile build, but to reduce the cognitive overload of parsing a large body of code, faced by developers in DevOps and other application developer scenarios, whose need is to get AI solutions integrated into their company’s infrastructure. 

A theme of this contribution is to create, from a monolithic body of code, a separation of concerns. A separation of concerns between the code to build servers from the code to build clients. To allow client developers to build and deploy simplified packages into their computing environment. To expedite their learning curve.

The use cases for building TensorFlow Serving are resource intensive and specialized. I’m not sure a makefile build adds much value to the serving use case.  Likely, no one is going to embed TensorFlow Serving into their smartphone app (Unlike TensorFlow itself).

TensorFlow Serving clients may be embedded throughout a computing environment and various applications. A clear separation of a body of client code is the main benefit. Any solution, mine or another, that creates this is valuable; including a solution that works within the Bazel build environment. However, a makefile build, in the client case, makes integration of TensorFlow AI solutions just that much more approachable to the people that are trying to make it happen.   

For anyone stumbling accross this closed Pull Requests, who may be interested in this solution, here is a link to the branch on my TensorFlow Serving fork.

This was developed as an intermediate step to facilitate developing specilized clients. I was happy to contribute it. 

https://github.com/mountaintom/serving/tree/create-makefile-serving-client/tensorflow_serving/contrib/makefile_client

The code pruned out of the main TensorFlow Serving codebase, in the above project, includes additional classes to support building the example Inception client. This is so a developer has a touchstone to check while building their own client and they can use the methods in the example without spending time exploring alternatives. 

In the process of building clients, I’ve had time to pull the lid off some of the supporting classes and look inside. In the client use case, the functionally provided by the Tensor class is trivial to implement. Further simplifying client builds where only the gRPC and Protocol Buffers generated objects need to be included should be possible.

chrisolston:
Thanks for the comment. To be clear, the issue on our side is more about the support burden and what we can realistically take on with the resources we have. We are not making a statement about the utility of this contribution. At this time we can only support bazel build infra (understood that it's an extra hurdle for some users, but it's something we can more readily support on our side at this time).

PR #468, opened 2017-06-06
FIx service selector

xuchenCN:
Fix the Selector  section of Service to the Deployment


xuchenCN:
I signed it!

kirilg:
Seems to be similar to https://github.com/tensorflow/serving/pull/442 but that PR didn't need to modify the name. Will that work as well?

kirilg:
Thanks for the fix and for following up on the other PR. Looks good, I'll merge.

PR #502, opened 2017-06-26
Add build for CPU

cheyang:
When building serving with [docker file](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/Dockerfile.devel), I noticed only bazel is built but not serving binary. So I add this piece to the docker file. Please review.


cheyang:
@kirilg , can you help take a look at this? Thanks.

chrisolston:
Thanks for the change!

PR #504, opened 2017-06-27
Add srcs_version = "PY2AND3" in BUILD files

sefira:
Hi all~

I've used tf to train my model with Anaconda3 (python 3.6), and I want to deploy it using this tf-serving.
But when I build it with ```--force_python=py3 ```, it seems some examples are missing the ```srcs_version = "PY2AND3"``` annotation in BUILD files.

I've succeeded in building tf-serving and runing with python 3.6 after adding these annotations.


sefira:
I signed it!

nkconnor:
I also ran into this. With the `tensorflow/models` dependency I had to alter the BUILD files in that repository to get it working

chrisolston:
Thanks for the change!

PR #506, opened 2017-06-27
add wget to list of apps to be installed

pankajkgupta:
The imagenet training script uses wget to download datasets and fails since it cannot find wget command in the docker container. Hence adding wget to the list would solve it.


pankajkgupta:
I signed the CLA!

chrisolston:
Thanks for the change!

Issue #530, opened 2017-07-21
Padding for variable-length inputs in batching session. 


penguin138:
Hi, I encountered a problem when using serving with RNNs that accept variable-length input. When passing batch of inputs of variable length to RNN I perform zero padding and save original sequence lengths, so that input tensor has fixed dimensions. When using serving with batching I am unable to do that, because apparently batching session does not perform padding (even optional) for batch of incoming requests. So when I send requests with inputs of different length server crashes when concatenating inputs in MergeInputTensors method of BatchingSession. This happens because Concat assumes that all tensors have identical dim sizes for all dims except zero ( it simply uses `tensors[0].shape` with zero dim size set to sum of zero dim sizes of all tensors), but it's not true for non-padded inputs to RNN. 
Here's full output:
```
...2017-07-21 11:27:52.365101: I tensorflow_serving/model_servers/main.cc:151] Building single TensorFlow model file config:  model_name: default model_base_path: /tinkoff/tfserving_ckpts/ model_version_policy: 0
2017-07-21 11:27:52.365333: I tensorflow_serving/model_servers/server_core.cc:375] Adding/updating models.
2017-07-21 11:27:52.365354: I tensorflow_serving/model_servers/server_core.cc:421]  (Re-)adding model: default
2017-07-21 11:27:52.465527: I tensorflow_serving/core/basic_manager.cc:698] Successfully reserved resources to load servable {name: default version: 1}
2017-07-21 11:27:52.465547: I tensorflow_serving/core/loader_harness.cc:66] Approving load for servable version {name: default version: 1}
2017-07-21 11:27:52.465555: I tensorflow_serving/core/loader_harness.cc:74] Loading servable version {name: default version: 1}
2017-07-21 11:27:52.465570: I external/org_tensorflow/tensorflow/contrib/session_bundle/bundle_shim.cc:360] Attempting to load native SavedModelBundle in bundle-shim from: /tinkoff/tfserving_ckpts/1
2017-07-21 11:27:52.465578: I external/org_tensorflow/tensorflow/cc/saved_model/loader.cc:236] Loading SavedModel from: /tinkoff/tfserving_ckpts/1
2017-07-21 11:27:52.508828: W external/org_tensorflow/tensorflow/core/platform/cpu_feature_guard.cc:45] The TensorFlow library wasn't compiled to use SSE4.1 instructions, but these are available on your machine and could speed up CPU computations.
[...]

PR #535, opened 2017-07-26
Add padding for batching session

penguin138:
PR for issue #530  
Currently padding is not optional and works for all dims of all tensors in input. I can make it optional, but need guidance on how to proceed (probably would need to change classes that use BatchingSession).


penguin138:
@chrisolston, as I explained in issue #530 you will have the same problem in BatchOp in tensorflow. So you might need to add something similar to this padding to ProcessBatch callback as well.

tensorflow_serving/batching/batching_session.cc
```diff
@@ -29,6 +31,7 @@ limitations under the License.
 #include "tensorflow_serving/servables/tensorflow/serving_session.h"
 #include "tensorflow_serving/util/cleanup.h"
 #include "tensorflow_serving/util/hash.h"
+#include "third_party/eigen3/unsupported/Eigen/CXX11/Tensor"
```
chrisolston: Hmmm. I'm not aware of any other place in the tf-serving codebase where we import eigen directly. Are you sure what you want to do can't be accomplished via the tensorflow layer? (If needed we could potentially add a utility in tensorflow.)

tensorflow_serving/batching/batching_session.cc
```diff
@@ -102,6 +105,127 @@ TensorSignature TensorSignatureFromSignatureDefs(
   return tensor_signature;
 }
 
+template<int num_dims>
+Eigen::array<std::pair<int32, int32>, num_dims> CreatePadding(Tensor tensor,
```
chrisolston: Would be better to use the Tensor abstraction, rather than the underlying eigen one, if that's possible.

tensorflow_serving/batching/batching_session.cc
```diff
@@ -102,6 +105,127 @@ TensorSignature TensorSignatureFromSignatureDefs(
   return tensor_signature;
 }
 
+template<int num_dims>
```
chrisolston: Please be sure every function, class & method has a comment explaining briefly what it does.

tensorflow_serving/batching/batching_session.cc
```diff
@@ -102,6 +105,127 @@ TensorSignature TensorSignatureFromSignatureDefs(
...
+Eigen::array<std::pair<int32, int32>, num_dims> CreatePadding(Tensor tensor,
+        const std::vector<int>& max_dim_sizes) {
+    Eigen::array<std::pair<int32, int32>, num_dims> padding;
+    for (unsigned int i = 0; i < max_dim_sizes.size(); ++i) {
+        if (max_dim_sizes[i] - tensor.dim_size(i) > 0) {
+            padding[i] = std::make_pair(0,
+                                        max_dim_sizes[i] - tensor.dim_size(i));
+        } else {
+            padding[i] = std::make_pair(0, 0);
```
chrisolston: nit: use the more modern { } instead of std::make_pair

tensorflow_serving/batching/batching_session.cc
```diff
@@ -102,6 +105,127 @@ TensorSignature TensorSignatureFromSignatureDefs(
...
+        }
+    }
+    return padding;
+}
+
+template <typename T, int num_dims>
+class PadTensor {
```
chrisolston: nit: class -> struct (and drop the public:)

tensorflow_serving/batching/batching_session.cc
```diff
@@ -102,6 +105,127 @@ TensorSignature TensorSignatureFromSignatureDefs(
...
+ public:
+    Tensor operator()(Tensor input,
+            Eigen::array<std::pair<int32, int32>, num_dims> padding) {
```
chrisolston: nit: Please avoid using std::pair. Define a struct instead with named elements.

tensorflow_serving/batching/batching_session.cc
```diff
@@ -102,6 +105,127 @@ TensorSignature TensorSignatureFromSignatureDefs(
...
+        }
+        case 6: {
+            Eigen::array<std::pair<int32, int32>, 6> padding;
+            padding = CreatePadding<6>(tensor, max_dim_sizes[tensor_name]);
+            PadTensor<T, 6> padding_functor = PadTensor<T, 6>();
+            new_tensor = padding_functor(tensor, padding);
+            break;
+        }
+        default:
+        // only ranks from 0 to 6 are supported
```
chrisolston: Please arrange to return an error then, instead of just not doing the padding.

Also, please document this limitation in the header file.

tensorflow_serving/batching/batching_session.cc
```diff
@@ -20,6 +20,8 @@ limitations under the License.
 #include "tensorflow/core/framework/tensor.h"
```
chrisolston: Naturally we'll want (1) documentation in the header file, and (2) test coverage. (I gather you want feedback on the basic approach before going all the way with those aspects, which makes sense.)

tensorflow_serving/batching/batching_session.cc
```diff
@@ -340,19 +464,59 @@ Status BatchingSession::MergeInputTensors(
 
   // For each input tensor name, a vector of tensors from the individual tasks.
   std::map<string, std::vector<Tensor>> tensors_to_merge;
-
```
chrisolston: This is a lot of code to in-line in this method. It would be nice to preserve the non-padding code if possible. Would it be possible to make a surgical AddPadding() helper method/function?

Also, can you add a "bool pad_variable_length_inputs" or something along those lines to BatchingSessionOptions.

I'm envisioning something that looks like:
... do some stuff
if (options_.pad_variable_length_inputs) {
  AddPadding(...);
}
... do more stuff

Is that doable?

penguin138:
@chrisolston, I tried to address all your comments, added tests and documentation. Please, review changes. Also, I have a question about tests. I added tests for individual padding functions, but I feel that I need to test how batching session works with that padding(create BatchingSession with pad_variable_length_inputs=true) . For that I need some simple model, that can accept tensors of  rank > 1, so that I could pad them. However, all BatchingSession tests use HalfPlusTwo model (because its SavedModel is already in Tensorflow repo), that only accepts float scalars. Is this ok if I add another model for tests to tfserving repo?

tensorflow_serving/batching/batching_session.cc
penguin138: I removed most of the padding code from MergeInputTensors.  Now I am only precomputing `max_dim_sizes` and call AddPadding (as you suggested) on tensor when it is about to be added to `tensors_to_merge`.

chrisolston:
@penguin123:
 1. Did you see my question at https://github.com/tensorflow/serving/issues/530#issuecomment-319220438 about whether we can get away without integrating padding into BatchingSession?
 2. re: end-to-end testing, perhaps you can add a new SavedModel that has padding under tensorflow_serving/batching/testdata
 3. re: testing individual padding functions: testing these individually is a good idea. it would be great if you could move them to a separate library (tensorflow_serving/batching/batching_util.h/cc) because (a) then the methods are public and are appropriate for unit tests, and (b) as you mentioned we may ultimately want to use them in the batch op in addition to BatchingSession.

penguin138:
@chrisolston, I answered the question at #530. Will add end-to-end tests later.

penguin138:
@chrisolston,  I added end-to-end test for padding in batching session and moved padding code to batching_util. Please, review changes.

tensorflow_serving/batching/BUILD
```diff
@@ -114,3 +121,28 @@ cc_test(
         "@org_tensorflow//tensorflow/core:test",
     ],
 )
+
+cc_library(
+    name = "batching_util",
+    srcs = ["batching_util.cc"],
+    hdrs = ["batching_util.h"],
+    visibility = ["//visibility:public"],
```
chrisolston: Please don't make it public. If you simply remove this line it will use the default internal visibility.

tensorflow_serving/batching/BUILD
```diff
@@ -114,3 +121,28 @@ cc_test(
...
+    deps = [
+        "@org_tensorflow//tensorflow/core:framework",
+        "@org_tensorflow//tensorflow/core:lib",
+        "@org_tensorflow//tensorflow/contrib/batching:basic_batch_scheduler",
```
chrisolston: I haven't looked inside yet but it doesn't seem right that batching-util should depend on basic-batch-scheduler. What's the reason?

tensorflow_serving/batching/batching_session.h
```diff
@@ -99,6 +99,23 @@ struct BatchingSessionOptions {
...
+  // they will be padded to shapes [1, 500, 101], [2, 500, 101], [1, 500, 101].
+  //
+  // This option is useful when using recurrent models(like LSTMs) with serving.
+  // These models typically accept variable-length inputs and when
+  // training them typical strategy is to save sequence lengths for decoding
+  // and pad those variable-length dims to maximum in batch.
+  // So, this flag is used to achieve similar behavior
+  // when serving with batching, it is assumed that sequence lengths
+  // have already been saved.
+  bool pad_variable_length_inputs;
```
chrisolston: add "= false" so that it's off by default (versus undefined)

tensorflow_serving/servables/tensorflow/session_bundle_config.proto
```diff
@@ -86,4 +86,7 @@ message BatchingParameters {
   //  - The entries must be in increasing order.
   //  - The final entry must equal 'max_batch_size'.
   repeated int64 allowed_batch_sizes = 6;
+
+  // Whether to pad variable-length inputs when batch is formed.
```
chrisolston: nit: "when a batch is formed"

tensorflow_serving/batching/batching_session.cc
```diff
@@ -79,6 +82,17 @@ TensorSignature TensorSignatureFromRunArgs(
   return signature;
 }
 
+std::vector<std::vector<std::pair<string, Tensor>>> GetTaskInputsVector(
```
chrisolston: Please ensure that all functions & methods have a comment explaining (at least briefly) what they do.

tensorflow_serving/batching/batching_session.cc
```diff
@@ -340,7 +354,12 @@ Status BatchingSession::MergeInputTensors(
 
   // For each input tensor name, a vector of tensors from the individual tasks.
   std::map<string, std::vector<Tensor>> tensors_to_merge;
-
+  std::map<string, std::vector<int>> max_dim_sizes;
```
chrisolston: Can you add a comment explaining what the map keys and values represent.

tensorflow_serving/batching/batching_session.h
```diff
@@ -99,6 +99,23 @@ struct BatchingSessionOptions {
...
+  //
+  // For example:
+  // given input tensors of shapes [1, 500, 101], [2, 300, 101], [1, 400, 101]
+  // they will be padded to shapes [1, 500, 101], [2, 500, 101], [1, 500, 101].
+  //
+  // This option is useful when using recurrent models(like LSTMs) with serving.
+  // These models typically accept variable-length inputs and when
+  // training them typical strategy is to save sequence lengths for decoding
+  // and pad those variable-length dims to maximum in batch.
+  // So, this flag is used to achieve similar behavior
```
chrisolston: nit: flag -> option

tensorflow_serving/batching/batching_session.h
```diff
@@ -99,6 +99,23 @@ struct BatchingSessionOptions {
   //
   // If left empty, no rounding/padding is performed.
   std::vector<int> allowed_batch_sizes;
+
+  // If set to true, padding is performed for tensors of the same name
+  // but with unequal dimensions, so that all tensors of the same name have
+  // equal dim sizes.
```
chrisolston: Please explain what data is used for the padding, and what data types are supported.

For numerical types I believe you pad with zeros.

For non-numerical types (e.g. string; are there any others?) what do you do -- simply not supported? Or pad with empty strings?

For the padding we do for allowed_batch_sizes we use existing values (not zeros) as the padding, because we know that these are legal values (or if not, the processing will fail for reasons not having to do with the padding). Perhaps we should do the same thing for this padding? WDYT?

tensorflow_serving/batching/batching_util.h
```diff
@@ -0,0 +1,206 @@
...
+// paddings[i].first - number of objects to add before elements in dimension i
+// of given tensor,
+// paddings[i].second - number of objects to add after elements in dimension i.
+// This padding signature is used when performing internal padding with Eigen.
+//
+// When building paddings it is assumed that tensor needs to be padded
+// after each dimension, so its shape matches max_dim_sizes,
+// except zeroth dimension which is left as is.
+template<int num_dims>
+std::array<std::pair<int32, int32>, num_dims> CreatePadding(Tensor tensor,
```
chrisolston: Please define a struct rather than using std::pair.

chrisolston: Any reason to use std::array instead of std::vector? If there isn't a specific good reason I'd rather use vector for uniformity with the rest of the code-base.

tensorflow_serving/batching/batching_util.h
```diff
@@ -0,0 +1,206 @@
...
+
+// Pads tensor so that its shape becomes as specified in max_dim_sizes,
+// Its zeroth dimension is unchanged.
+//
+// For example given tensor with shape [1, 2, 3, 4] and max_dim_sizes
+// [0, 2, 5, 8] function returns PaddingResult with padded_tensor of shape
+// [1, 2, 5, 8].
+//
+// Only tensors with rank from 0 to 6 are supported.
+PaddingResult AddPadding(const Tensor& tensor,
```
chrisolston: For consistency with our other function signatures, please change to:
Status AddPadding(..., Tensor* result)

(And eliminate the PaddingResult struct.)

tensorflow_serving/batching/batching_util.h
```diff
@@ -0,0 +1,206 @@
+/* Copyright 2017 Google Inc. All Rights Reserved.
```
chrisolston: The only function that needs to be exposed in the API here is AddPadding() right? Can you move the others to the .cc file and out of the public interface?

As part of this, the unit test would ideally only test AddPadding() (and cover the others indirectly).

tensorflow_serving/batching/batching_util.h
```diff
@@ -0,0 +1,206 @@
...
+// task2: {'features': [100, 200, 123], 'true_labels': [100]}
+// task3: {'features': [200, 100, 400], 'true_labels': [200]}
+//
+// the following map will be generated:
+// {'features': [0, 500, 400], 'true_labels': [0]}
+std::map<string, std::vector<int>> CalculateMaxDimSizes(
+    const std::vector<std::vector<std::pair<string, Tensor>>>& batch);
+
+// Pads tensor so that its shape becomes as specified in max_dim_sizes,
+// Its zeroth dimension is unchanged.
```
chrisolston: So does max_dim_sizes start with the first (not zeroth) dimension, or is the first entry in max_dim_sizes ignored?

tensorflow_serving/batching/batching_util.h
```diff
@@ -0,0 +1,206 @@
...
+// task1: {'features': [100, 500, 300], 'true_labels': [100]}
+// task2: {'features': [100, 200, 123], 'true_labels': [100]}
+// task3: {'features': [200, 100, 400], 'true_labels': [200]}
+//
+// the following map will be generated:
+// {'features': [0, 500, 400], 'true_labels': [0]}
+std::map<string, std::vector<int>> CalculateMaxDimSizes(
+    const std::vector<std::vector<std::pair<string, Tensor>>>& batch);
+
+// Pads tensor so that its shape becomes as specified in max_dim_sizes,
```
chrisolston: Please document what values are used for padding (see my other comment for discussion of what we want to do in terms of zeros or existing values).

tensorflow_serving/servables/tensorflow/session_bundle_config.proto
chrisolston: Thanks for threading the option through to here!

tensorflow_serving/batching/batching_session.h
```diff
@@ -99,6 +99,23 @@ struct BatchingSessionOptions {
...
+  //
+  // For example:
```
chrisolston: What will happen if a user tries to batch together tensor with different dimension sizes, and they have set pad_variable_length_inputs=false? They'll get an error Status right (and not a crash)? Please document this, and also ensure there's a unit test (perhaps there already is one).

tensorflow_serving/batching/batching_session.cc
```diff
@@ -20,6 +20,8 @@ limitations under the License.
 #include "tensorflow/core/framework/tensor.h"
 #include "tensorflow/core/framework/tensor_shape.h"
 #include "tensorflow/core/framework/tensor_util.h"
+#include "tensorflow/core/framework/types.h"
```
chrisolston: nit: I think you can remove these two #includes now that you've moved the type-specific handling to batching_util.

tensorflow_serving/batching/batching_session.cc
```diff
@@ -350,8 +369,15 @@ Status BatchingSession::MergeInputTensors(
       const Tensor& tensor = entry.second;
 
       std::vector<Tensor>& tensor_vec = tensors_to_merge[tensor_name];
-      tensor_vec.push_back(tensor);
-
+      Tensor new_tensor = tensor;
+      if (options_.pad_variable_length_inputs) {
+        PaddingResult res = AddPadding(tensor, max_dim_sizes[tensor_name]);
```
chrisolston: nit: res -> result (please avoid uncommon abbreviations in variable names)

tensorflow_serving/batching/batching_session.cc
```diff
@@ -350,8 +369,15 @@ Status BatchingSession::MergeInputTensors(
...
+        if (!res.padding_status.ok()) {
```
chrisolston: With the new AddPadding() signature I'm suggesting, you'll be able to use TF_RETURN_IF_ERROR.

tensorflow_serving/batching/batching_session.cc
chrisolston: Since you only populate/use this if padding is requested, can you make it an optional<map>, so it defaults to "no value set" rather than an empty map.

See tensorflow_serving/util/optional.h.

tensorflow_serving/batching/batching_session.cc
```diff
@@ -350,8 +369,15 @@ Status BatchingSession::MergeInputTensors(
       const Tensor& tensor = entry.second;
 
       std::vector<Tensor>& tensor_vec = tensors_to_merge[tensor_name];
-      tensor_vec.push_back(tensor);
-
+      Tensor new_tensor = tensor;
```
chrisolston: Instead of new_tensors can we do:

Tensor optionally_padded_tensor;
if (options_.pad_variable_length_inputs) {
  ...
  ... AddPadding(..., &optionally_padded_tensor);
} else {
  optionally_padded_tensor = tensor;
}
...

tensorflow_serving/batching/batching_util.h
penguin138: Well, I pass the value returned by this function to pad method of Eigen::TensorMap. When I try to pass std::vector instead of std::array it gives me compilation error saying "template instantiation depth exceeds maximum of 900" on [this line](https://github.com/RLovelett/eigen/blob/master/unsupported/Eigen/CXX11/src/util/CXX11Meta.h#L362).  There appears to be some recursive template array unfolding. It seems that Eigen needs to know size of array at compile time.

penguin138: I replaced std::pair with OneDimPadding struct, but I still needed to name its fields "first" and "second", because pad method of Eigen::TensorMap relies on template compile-time duck typing and accepts some container of objects that have fields "first" and "second" as padding signature. Btw, in PadOp in tensorflow they used std::pair.

tensorflow_serving/batching/batching_util.h
penguin138: Actually, if we pad to max length in batch then there are two methods: AddPadding and CalculateMaxDimSizes. I moved other padding functions to .cc and made changes to unit tests.

tensorflow_serving/batching/batching_session.h
penguin138: I now pad with the first element in tensor and support numeric, bool, string and quantized types.

penguin138:
@chrisolston, I addressed all your comments. Please, review.

tensorflow_serving/batching/BUILD
```diff
@@ -18,6 +18,12 @@ filegroup(
     ),
 )
 
+filegroup(
```
chrisolston: Shouldn't this go in batching/testdata/BUILD?

tensorflow_serving/batching/batching_session.cc
```diff
@@ -79,6 +81,34 @@ TensorSignature TensorSignatureFromRunArgs(
...
+  std::vector<std::vector<std::pair<string, Tensor>>> all_task_inputs;
+  for (int i = 0; i < batch.num_tasks(); ++i) {
+    const std::vector<std::pair<string, Tensor>>& task_inputs =
+      *batch.task(i).inputs;
+    all_task_inputs.push_back(task_inputs);
+  }
+  return all_task_inputs;
+}
+// Returns true iff all dims of shape1 are equal to dims of shape2 starting with
+// the first dimension.
```
chrisolston: Clarify that you are counting dimensions starting from zero, so maybe "starting with the first (not zeroth) dimension"

tensorflow_serving/batching/batching_session.cc
```diff
@@ -79,6 +81,34 @@ TensorSignature TensorSignatureFromRunArgs(
...
+// For example, for shapes [1, 2, 3] and [4, 2, 3] the result is true.
+bool CheckShapesEqualExceptZeroDim(const TensorShape& shape1,
```
chrisolston: nit: rename to AreShapesEqualExceptZeroDim() (since we sometimes use "check" to mean crash :)

tensorflow_serving/batching/batching_session.h
```diff
@@ -99,6 +99,33 @@ struct BatchingSessionOptions {
...
+  // given input tensors of shapes [1, 500, 101], [2, 300, 101], [1, 400, 101]
+  // they will be padded to shapes [1, 500, 101], [2, 500, 101], [1, 500, 101].
+  //
+  // Supported tensor datatypes:
+  // DT_FLOAT, DT_DOUBLE, DT_INT8, DT_UINT8, DT_INT16,
+  // DT_UINT16, DT_INT32, DT_INT64, DT_COMPLEX64, DT_COMPLEX128,
+  // DT_STRING, DT_BOOL, DT_QINT8, DT_QUINT8, DT_QINT16,
+  // DT_QUINT16, DT_QINT32, DT_HALF, DT_RESOURCE.
+
```
chrisolston: nit: add "//" on this line

tensorflow_serving/batching/batching_session.h
```diff
@@ -99,6 +99,33 @@ struct BatchingSessionOptions {
...
+  // If set to true, padding is performed for tensors of the same name
+  // but with unequal dimensions, so that all tensors of the same name have
+  // equal dim sizes.
+  // For each tensor its first element is used as padding value.
+  //
+  // For example:
+  // given input tensors of shapes [1, 500, 101], [2, 300, 101], [1, 400, 101]
+  // they will be padded to shapes [1, 500, 101], [2, 500, 101], [1, 500, 101].
+  //
+  // Supported tensor datatypes:
```
chrisolston: Are there any types *not* supported?

tensorflow_serving/batching/batching_session.h
```diff
@@ -99,6 +99,33 @@ struct BatchingSessionOptions {
   //
   // If left empty, no rounding/padding is performed.
   std::vector<int> allowed_batch_sizes;
+
+  // If set to true, padding is performed for tensors of the same name
+  // but with unequal dimensions, so that all tensors of the same name have
```
chrisolston: clarify: unequal dimensions **modulo the zeroth dimension**

tensorflow_serving/batching/batching_session_test.cc
```diff
@@ -181,6 +204,71 @@ TEST(BatchingSessionTest, Basic) {
...
+          }));
+  std::unique_ptr<Thread> second_request_thread(Env::Default()->StartThread(
+          ThreadOptions(), "second_request", [&batching_session] {
+              TestRequestToMatrixHalfPlusTwo({5, 6, 7, 8, 9, 10, 11, 12, 13},
+                  {1, 3, 3}, {4.5, 5, 5.5, 6, 6.5, 7, 7.5, 8, 8.5}, {1, 3, 3},
+                  batching_session.get());
+          }));
+}
+
+TEST(BatchingSessionTest, UnequalTensorShapesWithBatchingTurnedOff) {
```
chrisolston: BatchingTurnedOff -> PaddingTurnedOff

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+                               const std::vector<int>& max_dim_sizes,
+                               Tensor* output_tensor) {
+  int num_dims = tensor.dims();
+  Status padding_status;
+  switch (num_dims) {
+    case 0: {
+      std::array<OneDimPadding, 0> padding;
+      padding = CreatePadding<0>(tensor, max_dim_sizes);
+      PadTensor<T, 0> padding_functor = PadTensor<T, 0>();
+      padding_status = padding_functor(tensor, padding, output_tensor);
```
chrisolston: You can eliminate the padding_status variable and just in-line the returns, so:
return padding_functor(...);
instead of
padding_status = padding_functor(...);
break;
... and returning after the switch statement.

tensorflow_serving/batching/batching_util.h
```diff
@@ -0,0 +1,67 @@
...
+namespace tensorflow {
+namespace serving {
+
+// For batch of inputs calculates maximum dim sizes across all tensors
+// with the same name.
+// These dim sizes are used later to calculate padding amount for each tensor.
+// For example, for batch containing three tasks with the following inputs
+// (instead of tensors there are their shapes):
+//
+// task1: {'features': [100, 500, 300], 'true_labels': [100]}
```
chrisolston: I found "features" and "true_labels" confusing and distracting. Make it fully abstract and just use "tensor_a" and "tensor_b"?

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+#include "tensorflow/core/framework/tensor.h"
+#include "tensorflow/core/framework/types.h"
+#include "tensorflow/core/framework/register_types.h"
+#include "tensorflow/core/lib/core/errors.h"
+
+
+namespace tensorflow {
+namespace serving {
+
+// Padding for one dimension of a tensor.
```
chrisolston: I don't understand what this represents. What do "pad before" and "pad after" mean? Can you elaborate in the comments please.

Also, could "first" and "second" be renamed to "pad_before" and "pad_after"?

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+// of given tensor,
+// paddings[i].second - number of objects to add after elements in dimension i.
+// This padding signature is used when performing internal padding with Eigen.
+//
+// When building paddings it is assumed that tensor needs to be padded
+// after each dimension, so its shape matches max_dim_sizes,
+// First entry of max_dim_sizes, which is maximum size of zeroth dimension,
+// is ignored, because we don't perform padding in that dimension.
+template<int num_dims>
+std::array<OneDimPadding, num_dims> CreatePadding(Tensor tensor,
```
chrisolston: Prefer std::vector (for uniformity with the rest of the code) unless there's a specific good reason to use std::array here.

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+        const std::vector<int>& max_dim_sizes) {
+  std::array<OneDimPadding, num_dims> padding;
+  for (unsigned int i = 0; i < max_dim_sizes.size(); ++i) {
+    if (i > 0 && max_dim_sizes[i] - tensor.dim_size(i) > 0) {
+      padding[i] = {0, max_dim_sizes[i] - tensor.dim_size(i)};
```
chrisolston: Is pad_before always 0? If so could we simplify the code by dropping that and only having a single number (pad_after)?

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+    TensorShape output_shape;
+    for (int d = 0; d < num_dims; ++d) {
+      // Pad before existing elements.
+      const int32 before_d = padding[d].first;
+      // Pad after existing elements.
+      const int32 after_d = padding[d].second;
+      output_shape.AddDim(before_d + input.dim_size(d) + after_d);
+    }
+    if (output_shape.num_elements() == input.NumElements()) {
+      auto result = output->CopyFrom(input, output_shape);
```
chrisolston: Please avoid "auto" except for iterators (which have gnarly types that aren't usually helpful for readability of the code). Please find all non-iterator uses of "auto" in your change and put in the real time. Thanks!

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+      if (!result) {
+        return errors::Internal("Couldn't create output.");
+      }
+      return Status::OK();
+    }
+    *output = Tensor(input.dtype(), output_shape);
+    typename TTypes<T, num_dims>::Tensor inputs = input.tensor<T, num_dims>();
+    T pad_value(input.flat<T>()(0));  // using existing values in padding
```
chrisolston: Sorry if I missed it, but did you verify somewhere before this point that there is at least one element (i.e. input.flat<T>()(0) won't be a memory error)?

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+    output->tensor<T, num_dims>() = inputs.pad(padding, pad_value);
+    return Status::OK();
+  }
+};
+
+// functor specialization for scalars:
```
chrisolston: Do we need this? Does it even make sense to batch scalars? I'm not sure if the rest of the batching codebase supports scalars, so there may be no point in supporting them here. (But I could be wrong.)

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+// we do not perform padding in that case.
+template<typename T>
+struct PadTensor<T, 0> {
+  Status operator()(Tensor input,
+      const std::array<OneDimPadding, 0>&  padding, Tensor* output) {
```
chrisolston: Again, would std::vector work just as well as std::array here?

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+  for (int i = 0; i < batch.size(); ++i) {
+    const std::vector<std::pair<string, Tensor>>& task_inputs = batch[i];
+    for (const auto& entry : task_inputs) {
+      const string& tensor_name = entry.first;
+      const Tensor& tensor = entry.second;
+
+      std::vector<int>& max_dim_sizes_for_one_input =
+          max_dim_sizes[tensor_name];
+      for (int j = 0; j < tensor.dims(); ++j) {
+        int old_max_size = max_dim_sizes_for_one_input[j];
```
chrisolston: nit: const

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+  const std::vector<std::pair<string, Tensor>>& task_inputs = batch[0];
+  for (const auto& entry : task_inputs) {
+    const string& tensor_name = entry.first;
+    const Tensor& tensor = entry.second;
+    max_dim_sizes[tensor_name] = std::vector<int>(tensor.dims(), 0);
+  }
+  // fill
+  for (int i = 0; i < batch.size(); ++i) {
+    const std::vector<std::pair<string, Tensor>>& task_inputs = batch[i];
+    for (const auto& entry : task_inputs) {
```
chrisolston: (This is a good use of auto.)

tensorflow_serving/batching/batching_util.cc
```diff
@@ -0,0 +1,229 @@
...
+      }
+    }
+  }
+  return max_dim_sizes;
+}
+
+Status AddPadding(const Tensor& tensor,
+                  const std::vector<int>& max_dim_sizes,
+                  Tensor* padded_tensor) {
+  DataType input_dtype = tensor.dtype();
```
chrisolston: nit: const

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+
+#include <gtest/gtest.h>
+#include "tensorflow/core/framework/tensor.h"
+#include "tensorflow/core/lib/core/status.h"
+#include "tensorflow/core/platform/types.h"
+
+namespace tensorflow {
+namespace serving {
+namespace {
+std::vector<std::pair<string, Tensor>> CreateInputsWithTensorShapes(
```
chrisolston: Please add a comment explaining what this function does, and the format of the output.

(Please go through the entire PR and make sure all functions/methods are documented either in the .h file, or in the .cc file for ones not part of a .h file like this one. Thanks!)

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+  std::vector<std::pair<string, Tensor>> inputs;
+  for (int i = 0; i < shapes.size(); ++i) {
+    inputs.push_back({"x" + std::to_string(i), Tensor(DT_FLOAT, shapes[i])});
+  }
+  return inputs;
+}
+
+
+TEST(BatchingUtilTest, CalculateMaxDimSizes) {
+  std::vector<TensorShape> shapes1;
```
chrisolston: Rather than using distinct variable names (shapes1, shapes2, ...) it would be better to use scopes, as in:

{
  shapes = ...;
  EXPECT_...(shapes ...);
}
{
  shapes = ...;
  ...and so on
}

And similarly elsewhere in your test code please.

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+  shapes1.push_back({10, 20, 30});  // x0
```
chrisolston: Please use nested brace initialization instead of these push_backs to construct shapes1 etc.:
const std::vector<TensorShape> shapes1 = {{10,20,30}, {10,100}};

Similarly elsewhere in your test code please.

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+  shapes2.push_back({20, 101});  // x1
+  std::vector<std::pair<string, Tensor>> inputs2 =
+    CreateInputsWithTensorShapes(shapes2);
+  std::vector<std::vector<std::pair<string, Tensor>>> batch {inputs1, inputs2};
+  std::map<string, std::vector<int>> max_dim_sizes =
+    CalculateMaxDimSizes(batch);
+  std::map<string, std::vector<int>> true_max_dim_sizes;
+  true_max_dim_sizes["x0"] = {20, 50, 30};
+  true_max_dim_sizes["x1"] = {20, 101};
+  EXPECT_EQ(max_dim_sizes, true_max_dim_sizes);
```
chrisolston: Please use the gmock matchers instead (they give more readable error messages if the test fails):
EXPECT_THAT(max_dim_sizes, UnorderedElementsAre(Pair("x0", ElementsAre(20,50,30), ...

Same elsewhere in your test code please.

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+}
+
+TEST(BatchingUtilTest, AddPadding) {
+  std::vector<int> max_dim_sizes {20, 100, 200};
```
chrisolston: nit: const

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+  std::vector<DataType> types {DT_FLOAT, DT_DOUBLE, DT_INT32, DT_UINT8,
```
chrisolston: nit: const (in general, please try to make variables const when possible)

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+      DT_INT16, DT_UINT16, DT_INT8, DT_STRING, DT_COMPLEX64, DT_COMPLEX128,
+      DT_INT64, DT_BOOL, DT_QINT8, DT_QUINT8, DT_QINT16,
+      DT_QUINT16, DT_QINT32, DT_HALF, DT_RESOURCE};
+  Status padding_status;
+  for (auto type : types) {
+    Tensor tensor(type, {10, 20, 30});
+    Tensor padded_tensor;
+    padding_status = AddPadding(tensor, max_dim_sizes, &padded_tensor);
+    EXPECT_EQ(padding_status, Status::OK());
+    EXPECT_EQ(padded_tensor.shape(), TensorShape({10, 100, 200}));
```
chrisolston: nit: put the expected value *before* the actual one:
EXPECT_EQ(TensorShape(...) /* expected */, padded_tensor.shape() /* actual */);

Same elsewhere...

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+  EXPECT_EQ(max_dim_sizes, true_max_dim_sizes);
+}
+
+TEST(BatchingUtilTest, AddPadding) {
+  std::vector<int> max_dim_sizes {20, 100, 200};
+  std::vector<DataType> types {DT_FLOAT, DT_DOUBLE, DT_INT32, DT_UINT8,
+      DT_INT16, DT_UINT16, DT_INT8, DT_STRING, DT_COMPLEX64, DT_COMPLEX128,
+      DT_INT64, DT_BOOL, DT_QINT8, DT_QUINT8, DT_QINT16,
+      DT_QUINT16, DT_QINT32, DT_HALF, DT_RESOURCE};
+  Status padding_status;
```
chrisolston: Any reason not to inline the status check (eliminate "padding_status" variable), as in:

// (no padding_status variable)
for (...) {
  ...
  ASSERT_EQ(AddPadding(...));
  EXPECT_EQ(... shapes ...)
}

tensorflow_serving/batching/batching_util_test.cc
```diff
@@ -0,0 +1,93 @@
...
+  for (auto type : types) {
+    Tensor tensor(type, {10, 20, 30});
+    Tensor padded_tensor;
+    padding_status = AddPadding(tensor, max_dim_sizes, &padded_tensor);
+    EXPECT_EQ(padding_status, Status::OK());
```
chrisolston: Use ASSERT instead of EXPECT for things like status checks. ASSERT is for when it doesn't make sense to continue the test if it doesn't succeed, because something has gone off the rails.

tensorflow_serving/servables/tensorflow/session_bundle_config.proto
```diff
@@ -86,4 +86,7 @@ message BatchingParameters {
   //  - The entries must be in increasing order.
   //  - The final entry must equal 'max_batch_size'.
   repeated int64 allowed_batch_sizes = 6;
+
+  // Whether to pad variable-length inputs when a batch is formed.
+  google.protobuf.BoolValue pad_variable_length_inputs = 7;
```
chrisolston: Can we just use "bool" instead of BoolValue? We don't need three-valued logic here, i.e. just defaulting to "false" if it's omitted is fine right?

tensorflow_serving/batching/batching_util.cc
penguin138: pad_after is number of values to add before elements in one dimension, and pad_after is number of objects to add after. I will add that to comments. Unfortunately I needed to name fields "first" and "second", because pad method of Eigen::TensorMap (where I pass array of OneDimPaddings to) relies on template compile-time duck typing and accepts some container of objects that have fields "first" and "second" as padding signature. In PadOp they used std::pair for instance.

tensorflow_serving/batching/batching_util.cc
penguin138: Well, I pass the value returned by this function to pad method of Eigen::TensorMap. When I try to pass std::vector instead of std::array it gives me compilation error saying "template instantiation depth exceeds maximum of 900" on this line. There appears to be some recursive template array unfolding. It seems that Eigen needs to know size of array at compile time.

tensorflow_serving/batching/batching_util.cc
penguin138: As I said in the comment above, we can't drop it, because of the signature that pad method of Eigen::TensorMap accepts.

tensorflow_serving/batching/batching_session.h
penguin138: Well, there is DT_BFLOAT, which is as I understand (I might be wrong) is only used internally and DT_VARIANT which was added to tensorflow about a month ago. Tfserving uses earlier version of tensorflow, so I can't add support for DT_VARIANT.

tensorflow_serving/batching/batching_util_test.cc
penguin138: As I understand scopes are useful when there are multiple asserts, but in this case inputs1 and inputs2 are used to form a batch that is used in only one assert. So, I think I can't  use scopes in this test case. If I find other cases where I can use scopes I'll do that.

penguin138:
@chrisolston, I fixed almost all code style issues, for those I didn't fix I left a comment. Please, review.

chrisolston:
Thanks for the contribution @penguin138. It's great that you helped us fix a gap in our offering. And thanks for your patience in working with me to get the code in shape to submit. I hope this will unblock your deployment.

dylanrandle:
@penguin138 You are an awesome person

dylanrandle:
Hi,

I am trying to run 3 models with padding turned on. I get the message:
```
Set pad_variable_length_inputs to true
```

How do I do that? Sorry for the easy question.

Thanks

chrisolston:
Set it via the BatchingParameters protobuf message:
https://github.com/tensorflow/serving/blob/master/tensorflow_serving/servables/tensorflow/session_bundle_config.proto#L101

You pass in a BatchingParameters message via a small text file with the flag --batching_parameters_file

dylanrandle:
Thanks @chrisolston 
For anyone wondering, add this line to your batching_parameters_file:
```pad_variable_length_inputs: true ```
Cheers

PR #539, opened 2017-07-28
Upstream changes

kirilg:


tensorflow_serving/model_servers/BUILD
```diff
@@ -186,6 +186,7 @@ pkg_tar(
     package_dir = "/usr/bin",
 )
 
+# Build with '-c opt --copt=-mavx --copt=-msse4.2'
```
promiseofcake: This this specified like this because of issues with march=native not pulling the correct instructions?

kirilg: This is to build a Debian package with these architecture specific optimizations that would have to broadest reach (most modern CPUs have support for these). The target above is compiled with just '-c opt' for even broader reach (everyone) but not as performant.

Issue #540, opened 2017-07-28
GPU Docker Fails to Find libcudnn


mhamilton723:
Ubuntu 16.06 Nvidia K80 GPU

When I run the gpu docker image with 

docker build --pull -t $USER/tensorflow-serving-devel-gpu -f tensorflow_serving/tools/docker/Dockerfile.devel-gpu .

It fails at building tensorflow and just repeats the following:

"Please specify the cuDNN version you want to use. [Leave empty to default to cuDNN 6.0]: Please specify the location where cuDNN 6 library is installed. Refer to README.md for more details. [Default is /usr/local/cuda]:Invalid path to cuDNN 6 toolkit. None of the following files can be found:
/usr/local/cuda-8.0/lib64/libcudnn.so.6
/usr/local/cuda-8.0/libcudnn.so.6
/usr/lib/x86_64-linux-gnu/libcudnn.so.6
"Please specify the cuDNN version you want to use. [Leave empty to default to cuDNN 6.0]: Please specify the location where cuDNN 6 library is installed. Refer to README.md for more details. [Default is /usr/local/cuda]:Invalid path to cuDNN 6 toolkit. None of the following files can be found:
/usr/local/cuda-8.0/lib64/libcudnn.so.6
/usr/local/cuda-8.0/libcudnn.so.6
/usr/lib/x86_64-linux-gnu/libcudnn.so.6

mhamilton723:
After investigating inside the docker image it appears that tensorflow defaults to use cudnn6 instead of cudnn5. Working on a fix now but if anyone has tips please post them here

kirilg:
The PR was merged, so I'm closing this issue.

Issue #327, opened 2017-02-19
bazel GPU build error with fatal error: external/nccl_archive/src/nccl.h: No such file or directory


cheyang:
We are trying to build Tensorflow Serving 0.5.1 with [TensorFlow 1.0.0@07bb8ea](https://github.com/tensorflow/tensorflow/commit/07bb8ea2379bd459832b23951fb20ec47f3fdbd4)

Basing on CUDA 7.5, cuDNN 5.
Bazel 0.4.4

```
cd serving && bazel build -c opt --config=cuda tensorflow_serving/...
```

```
ERROR: /root/.cache/bazel/_bazel_root/f8d1071c69ea316497c31e40fe0160
8c/external/org_tensorflow/tensorflow/contrib/nccl/BUILD:23:1: C++ c
ompilation of rule '@org_tensorflow//tensorflow/contrib/nccl:python/
ops/_nccl_ops.so' failed: crosstool_wrapper_driver_is_not_gcc failed
: error executing command external/local_config_cuda/crosstool/clang
/bin/crosstool_wrapper_driver_is_not_gcc -U_FORTIFY_SOURCE '-D_FORTI
FY_SOURCE=1' -fstack-protector -fPIE -Wall -Wunused-but-set-paramete
r ... (remaining 76 argument(s) skipped): com.google.devtools.build.
lib.shell.BadExitStatusException: Process exited with status 1.
In file included from external/org_tensorflow/tensorflow/contrib/ncc
l/kernels/nccl_manager.cc:15:0:
external/org_tensorflow/tensorflow/contrib/nccl/kernels/nccl_manager
.h:23:44: fatal error: external/nccl_archive/src/nccl.h: No such fil
e or directory
 #include "external/nccl_archive/src/nccl.h"
                                            ^
compilation terminated.
INFO: Elapsed time: 147.378s, Critical Path: 107.11s
```

I'm able to find nccl.h, but it can't be found during bazel build. Any suggestions? Thanks in advanced.
```
find / -name nccl.h
/root/.cache/bazel/_bazel_root/5071e8dca1385fb776f72b33971bf157/exte
rnal/nccl_archive/src/nccl.h
/root/.cache/bazel/_bazel_root/f8d1071c69ea316497c31e40fe01608c/exte
rnal/nccl_archive/src/nccl.h
```

tvkpz:
Same error here.

[...]

PR #541, opened 2017-07-28
Update gpu docker file

mhamilton723:
The Gpu Docker does not build on master (#540) for several reasons 

1) Tensorflow defaults to cudadnn_6.0
2) The issue raised and solved in 
https://github.com/tensorflow/serving/issues/327#issuecomment-282207825

This PR solves these two issues and the resulting docker image builds and passes all


mhamilton723:
I signed it!

mhamilton723:
What are the thoughts of the tf serving team?

kirilg:
Thanks for the contribution Mark!

Issue #613, opened 2017-10-05
TF server platform config file ignores malformed source_adapter_config


ales-t:
The field `source_adapter_config` has type `Any`. If the serialized value is malformed, the error is silently ignored. I'm not quite sure but I suspect the problem is on this line:

https://github.com/tensorflow/serving/blob/6330edb4bb7002b3bf8d32860c2e7fb0d5ab0a16/tensorflow_serving/util/class_registration.h#L269

AFAICT the method `google::protobuf::Any::UnpackTo` does not throw exceptions but instead returns a Boolean. The return value is ignored on the referenced line.

chrisolston:
Sorry for the bug. I'll be happy to review a PR that fixes it.

chrisolston:
It looks like the PR is all set. I'm waiting for you to sign the CLA before I can accept it.

PR #614, opened 2017-10-06
check for UnpackTo return value when parsing platform_config file

ales-t:
Fixes issue https://github.com/tensorflow/serving/issues/613

### DIFF
--- a/tensorflow_serving/util/class_registration.h
+++ b/tensorflow_serving/util/class_registration.h
@@ -267,7 +267,8 @@ class ClassRegistry {
             ->GetPrototype(descriptor)
             ->New());
     if (!any_config.UnpackTo(config.get())) {
-      return errors::InvalidArgument("Malformed content of Any.value");
+      return errors::InvalidArgument("Malformed content of Any.value: ",
+          any_config.DebugString());
     }
     return Create(*config, std::forward<AdditionalFactoryArgs>(args)...,
                   result);
--- a/tensorflow_serving/util/class_registration.h
+++ b/tensorflow_serving/util/class_registration.h
@@ -266,7 +266,9 @@ class ClassRegistry {
         protobuf::MessageFactory::generated_factory()
             ->GetPrototype(descriptor)
             ->New());
-    any_config.UnpackTo(config.get());
+    if (!any_config.UnpackTo(config.get())) {
+      return errors::InvalidArgument("Malformed content of Any.value");
+    }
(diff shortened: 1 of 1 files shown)


tensorflow_serving/util/class_registration.h
```diff
@@ -266,7 +266,9 @@ class ClassRegistry {
         protobuf::MessageFactory::generated_factory()
             ->GetPrototype(descriptor)
             ->New());
-    any_config.UnpackTo(config.get());
+    if (!any_config.UnpackTo(config.get())) {
+      return errors::InvalidArgument("Malformed content of Any.value");
```
chrisolston: It would be helpful to print the malformed thing here, if possible. Would something like this work:

return errors::InvalidArgument(...) << any_config.DebugString();

chrisolston:
Jenkins, test this please.

tensorflow_serving/util/class_registration.h
ales-t: Thanks, I changed the error message to include the `DebugString`.

chrisolston:
Jenkins, test this please.

tensorflow_serving/util/class_registration.h
```diff
@@ -266,7 +266,10 @@ class ClassRegistry {
         protobuf::MessageFactory::generated_factory()
             ->GetPrototype(descriptor)
             ->New());
-    any_config.UnpackTo(config.get());
+    if (!any_config.UnpackTo(config.get())) {
+      return errors::InvalidArgument("Malformed content of Any.value: ",
```
chrisolston: How do we know that false means there was a problem with the .value component specifically?

Maybe s/Malformed content of Any.value/Malformed content of Any/

ales-t: I'd say we know: Any only has two fields, `type_id` and `value`. `type_id` is resolved and the corresponding proto descriptor is loaded at this point.

chrisolston: Well, the internal representation of the Any could, in principle, change over time. But it's not a big deal either way.

chrisolston:
Thanks for the PR. Ping me when the CLA is set up and I'll merge it in.

tensorflow_serving/util/class_registration.h
ales-t: Good point, I changed the error message.

ales-t:
I signed it!

PR #616, opened 2017-10-10
Update tensorflow

vladmos:
The currently used version of Tensorflow contains an outdated version of protobuf which is not compatible with the upcoming Bazel 0.8.


kirilg:
Thanks for the PR! TF@head is not currently compatible with TF Serving, there's a known test failure that we're working to resolve and will update the submodule when it's fixed.

vladmos:
Do you have any estimations when the test will be fixed approximately? Do I need to keep this pull request or should I abandon it and you'll update the submodule yourself?

kirilg:
+ @case540 I don't have an estimate, but yes I think we'll want to update the submodule along with the fix the next time we push (the fix might be in TF so we'll need to sync past what's in this PR).

vladmos:
Is there a Github issue I can track?

kirilg:
The issue has been resolved and we synced to a newer TF.

Issue #227, opened 2016-10-28
Tensorflow Serving docker failed compilation


demiladef:
I have been trying to set tensorflow_serving up on a docker container for the past day or so with no luck.

I followed the instructions very carefully

`docker build --pull -t $USER/tensorflow-serving-devel -f Dockerfile.devel .`

`docker run -it $USER/tensorflow-serving-devel`

```
git clone --recurse-submodules https://github.com/tensorflow/serving
cd serving/tensorflow
./configure
cd ..
bazel test tensorflow_serving/...
```

But always end up here:

```
[4,197 / 6,007] Still waiting for 200 jobs to complete:
      Running (standalone):
        Compiling external/org_tensorflow/tensorflow/core/kernels/argmax_op.cc\
 [for host], 996 s
        Compiling external/org_tensorflow/tensorflow/core/kernels/batch_matmul\
_op_real.cc [for host], 992 s
        Compiling external/org_tensorflow/tensorflow/core/kernels/bias_op.cc [\
for host], 623 s
        Compiling external/org_tensorflow/tensorflow/core/kernels/cwise_op_gre\
ater.cc [for host], 130 s
        Compiling external/org_tensorflow/tensorflow/core/kernels/cwise_op_equ\
ERROR: /root/.cache/bazel/_bazel_root/f8d1071c69ea316497c31e40fe01608c/external/org_tensorflow/tensorflow/core/kernels/BUILD:1296:1: C++ compilation of rule '@org_tensorflow//tensorflow/core/kernels:cwise_op' failed: gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE '-D_FORTIFY_SOURCE=1' -fstack-protector -Wall -Wl,-z,-relro,-z,now -B/usr/bin -B/usr/bin -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g0 ... (remaining 103 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 4.
gcc: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-4.8/README.Bugs> for instructions.
INFO: Elapsed time: 2312.066s, Critical Path: 2111.50s
//tensorflow_serving/batching:basic_batch_scheduler_test              NO STATUS
//tensorflow_serving/batching:batch_scheduler_retrier_test            NO STATUS
[...]

PR #620, opened 2017-10-16
use bazel 0.5.4 in Dockerfile

Atlas7:
The bazel build step requires a minimum bazel version of `0.5.4`. The Dockerfile currently uses bazel `0.5.1` (which is too low). I've tested locally on my Macbook Pro - build ok with bazel `0.5.4` docker container.

To replicate test [follow instruction at](https://github.com/Atlas7/How-to-Deploy-a-Tensorflow-Model-in-Production/blob/mymaster/demo.ipynb) - but with bazel version `0.5.4`.

### DIFF
--- a/tensorflow_serving/tools/docker/Dockerfile.devel
+++ b/tensorflow_serving/tools/docker/Dockerfile.devel
@@ -34,7 +34,7 @@ RUN pip install mock grpcio
 
 ENV BAZELRC /root/.bazelrc
 # Install the most recent bazel release.
-ENV BAZEL_VERSION 0.5.1
+ENV BAZEL_VERSION 0.5.4
 WORKDIR /
 RUN mkdir /bazel && \
     cd /bazel && \
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -25,4 +25,4 @@ tf_serving_workspace()
 # Specify the minimum required bazel version.
 load("@org_tensorflow//tensorflow:workspace.bzl", "check_version")
 
-check_version("0.4.5")
+check_version("0.5.4")
--- a/tensorflow_serving/g3doc/setup.md
+++ b/tensorflow_serving/g3doc/setup.md
@@ -6,7 +6,7 @@ To compile and use TensorFlow Serving, you need to set up some prerequisites.
 
 ### Bazel (only if compiling source code)
 
-TensorFlow Serving requires Bazel 0.4.5 or higher. You can find the Bazel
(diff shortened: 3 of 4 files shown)


Atlas7:
Also briefly mentioned about this in [this thread](https://github.com/tensorflow/serving/issues/227)

tensorflow_serving/tools/docker/Dockerfile.devel
```diff
@@ -34,7 +34,7 @@ RUN pip install mock grpcio
 
 ENV BAZELRC /root/.bazelrc
 # Install the most recent bazel release.
-ENV BAZEL_VERSION 0.5.1
+ENV BAZEL_VERSION 0.5.4
```
kirilg: Thanks! Could you please update the version everywhere else as well:
https://github.com/tensorflow/serving/blob/828786c37a1f79bc1234a829efa1f0ca6d62da19/WORKSPACE#L28
tensorflow_serving/tools/docker/Dockerfile.devel-gpu
tensorflow_serving/g3doc/setup.md

Atlas7: No problem - on it :)

Atlas7: Done! Please check again. Should be all on bazel 0.5.4 now.  (Just one quick note - I notice the [most recent Bazel GitHub release](https://github.com/bazelbuild/bazel/releases) is actually 0.6.1. But as 0.5.4 is the minimal required from what I encountered and worked, I deliberately use 0.5.4 for now. This I guess will also avoid too many moving parts. My suggestion is to do this in phases. Get this all synced up to bazel 0.5.4 for the time being - knowing that it would work. At a later time we may upgrade to 0.6.x - when needed. (just to be safe / avoid too many moving parts)

promiseofcake:
We use 0.5.4 in production, looks good to me.

Atlas7:
Excellent! Thanks for the info :)

kirilg:
Jenkins, test this please.

Atlas7:
@kirilg Any idea why the `WORKSPACE` test failure? (I've tried to hunt down some clues from history but I guess I might know something that I don't? :)

kirilg:
Jenkins, test this please.

kirilg:
It was just a transient GitException, not an actual test failure. Reran and it now passed, so I'll merge. Thanks!

Atlas7:
Thank you so much! :)

PR #632, opened 2017-10-31
Adds a link to the referenced tutorial

sarahmaddox:
Adds a link to the MNIST For ML Beginners tutorial, which is currently referred to with text alone.


kirilg:
Thanks! Looks good.

PR #650, opened 2017-11-08
With your newly supplied code, I ran the bazel compile error。“config/BUILD:36:1: no such target '//tensorflow_serving/sources/storage_path:file_system_storage_path_source_proto_py_pb2_genproto': target 'file_system_storage_path_source_proto_py_pb2_genproto' not declared in package 'tensorflow_serving/sources/storage_path' defined by /home/tf/workspace/serving/tensorflow_serving/sources/storage_path/BUILD and referenced by '//tensorflow_serving/config:model_server_config_proto_py_pb2_genproto'”

tensorflower:


chrisolston:
Hi, can you please fix the PR name (it should say what code change the PR proposes to make).

And about that, it seems to be adding an RPC endpoint for reloading the model config; but what's the extra logging thing? Is that related, or could that be moved to a separate PR?

PR #654, opened 2017-11-13
Sync code example in docs with original source file

frreiss:
The example code copied from https://github.com/tensorflow/serving/blob/master/tensorflow_serving/example/mnist_saved_model.py to `g3doc/serving_basic.md` was out of date with a recent (~4 months ago) change in the original source file. This PR updates the code snippet in the Markdown file.


kirilg:
Thanks!

PR #666, opened 2017-11-20
Rewrote (simplified) "sources" presentation.

sanchom:
Unsure if I've changed the meaning. If I did, that was accidental. I found this section difficult to read and hope this is clearer.


kirilg:
Thanks for updating the documentation!

PR #694, opened 2017-12-14
add flag to config per_process_gpu_memory_fraction

zhouyoulie:
With "per_process_gpu_memory_fraction" flag (1-100), the users can config per_process_gpu_memory_fraction, otherwise the server process will occupy all the gpu memory.

### DIFF
--- a/tensorflow_serving/model_servers/main.cc
+++ b/tensorflow_serving/model_servers/main.cc
@@ -76,6 +76,7 @@ limitations under the License.
 #include "tensorflow_serving/servables/tensorflow/multi_inference.h"
 #include "tensorflow_serving/servables/tensorflow/predict_impl.h"
 #include "tensorflow_serving/servables/tensorflow/regression_service.h"
+#include "tensorflow_serving/servables/tensorflow/session_bundle_config.pb.h"
 
 namespace grpc {
 class ServerCompletionQueue;
@@ -302,6 +303,7 @@ tensorflow::serving::PlatformConfigMap ParsePlatformConfigMap(
 int main(int argc, char** argv) {
   tensorflow::int32 port = 8500;
   bool enable_batching = false;
+  tensorflow::int32 per_process_gpu_memory_fraction = 100;
   tensorflow::string batching_parameters_file;
   tensorflow::string model_name = "default";
   tensorflow::int32 file_system_poll_wait_seconds = 1;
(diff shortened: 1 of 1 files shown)


tensorflow_serving/model_servers/main.cc
```diff
@@ -346,7 +348,10 @@ int main(int argc, char** argv) {
                        "If non-empty, read an ascii PlatformConfigMap protobuf "
                        "from the supplied file name, and use that platform "
                        "config instead of the Tensorflow platform. (If used, "
-                       "--enable_batching is ignored.)")};
+                       "--enable_batching is ignored.)"),
+      tensorflow::Flag("per_process_gpu_memory_fraction", &per_process_gpu_memory_fraction,
+                       "per_process_gpu_memory_fraction in percent")};
```
chrisolston: Please use a floating point number (so from 0.0 to 1.0).

tensorflow_serving/model_servers/main.cc
```diff
@@ -302,6 +303,7 @@ tensorflow::serving::PlatformConfigMap ParsePlatformConfigMap(
 int main(int argc, char** argv) {
   tensorflow::int32 port = 8500;
   bool enable_batching = false;
+  tensorflow::int32 per_process_gpu_memory_fraction = 100;
```
chrisolston: Please have it default to 0, not 100. (0 is the default value in the tf config proto.)

tensorflow_serving/model_servers/main.cc
```diff
@@ -346,7 +348,10 @@ int main(int argc, char** argv) {
                        "If non-empty, read an ascii PlatformConfigMap protobuf "
                        "from the supplied file name, and use that platform "
                        "config instead of the Tensorflow platform. (If used, "
-                       "--enable_batching is ignored.)")};
+                       "--enable_batching is ignored.)"),
+      tensorflow::Flag("per_process_gpu_memory_fraction", &per_process_gpu_memory_fraction,
```
chrisolston: Please document that the default value is 0, which has special meaning and causes tensorflow to automatically select a value.

tensorflow_serving/model_servers/main.cc
zhouyoulie: have fixed~

tensorflow_serving/model_servers/main.cc
zhouyoulie: have fixed~

tensorflow_serving/model_servers/main.cc
zhouyoulie: have fixed~

chrisolston:
Thanks for the contribution.

zhouyoulie:
nice to learn from u～

PR #707, opened 2017-12-28
Use print function for python3 compatibility.

jkinkead:
This makes the examples compatible with python3 (and matches the client code in this directory, which already uses `print_function`).


kirilg:
Looks good, thanks!

kirilg:
Jenkins, test this please

PR #709, opened 2017-12-29
Initialize the variables within a session for inception saved model

bugra:
This PR introduces the following changes:
  - Initialize variables within a session before building and saving the model

This is to prevent `FailedPreconditionError`:
```
FailedPreconditionError: Attempting to use uninitialized value aux_logits/Conv/BatchNorm/beta
```

Tensorflow and Tensorflow Board versions to reproduce:
```
tensorflow==1.4.1
tensorflow-tensorboard==0.4.0rc3
```

### DIFF
--- a/tensorflow_serving/example/inception_saved_model.py
+++ b/tensorflow_serving/example/inception_saved_model.py
@@ -60,6 +60,8 @@ def export():
       texts[parts[0]] = parts[1]
 
   with tf.Graph().as_default():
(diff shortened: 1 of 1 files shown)


gautamvasudevan:
Hey bugra, I haven't been able to get this to repro and just went through the tutorial. Mind first taking a pass with the latest and filing an issue with repro steps? Thanks!

Issue #422, opened 2017-04-26
Best practice: loading new models during runtime (not versions)


markusnagel:
I'm curious to see/discuss what are the best options to deploy new models during runtime. I learned from the tutorials and examples how new versions of a model are loaded as well as how to load multiple models (at the start of the server). Unfortunately I could not find the functionality of loading new models during runtime. (I'm aware that parts where discussed in this issue #380.) I guess my first question is if there exists already any type of dynamic loading of new models during runtime? (I think technically all components should be there, however I could not find a 'trigger' which starts the process of loading the model.)

In general I can think of several ways to serve multiple models during runtime:
- Use a separate server for each model. A small manager module could start a new server or even docker container for each model. (Question: How to best forward the prediction request to the right server?)
- Similar to watching multiple versions on the file system, we could watch the parent directory if a new folder for a new model.
- If we use the 'custom_model_config' we periodically check it for changes and see if a new model is in there (or one is missing).
- Have a gRPC call which sends the required information for the new model.

It would be great if some of the developers could post their opinion on how such a case should be handled or people share their opinion on best practices related to deploying new models.

chrisolston:
Probably the best would be to extend model_servers/main.cc to accept new ModelServerConfigs (perhaps via an rpc call), and call ServerCore::ReloadConfig().

rajhans:
I am also working on dynamically loading new models. Is this -- i.e. using RPC to ReloadConfig -- still the recommended route for adding new models? 

If so, one question: when we call ReloadConfig, is there a downtime when the models get reloaded or are the models replaced only when new models are loaded and ready to serve?

chrisolston:
Correct, that is still the recommended approach.

During ReloadConfig() any already-loaded models will remain loaded and can take traffic the whole time.

rajhans:
I have managed to get it working by adding a Reload endpoint to PredictionServiceImpl in main.cc. Would you be interested in getting these changes as a pull request?

chrisolston:
@rajhans - That sounds good, but can you please define a separate service for this new RPC endpoint, rather than adding it to PredictionService?

Also, it would be good for the new config to be passed in-line in the rpc request, rather than having the rpc pass a filename for the config.

[...]

PR #718, opened 2018-01-08
[Awaiting early feedback]Timeout variable to ReloadConfig calls

rajhans:
Based on the discussion in #422.

### DIFF
--- a/tensorflow_serving/core/servable_state_monitor.cc
+++ b/tensorflow_serving/core/servable_state_monitor.cc
@@ -197,7 +197,7 @@ bool ServableStateMonitor::WaitUntilServablesReachState(
     const std::vector<ServableRequest>& servables,
     const ServableState::ManagerState goal_state,
     std::map<ServableId, ServableState::ManagerState>* const states_reached,
-    const int timeout) {
+    const int64 timeout_in_us) {
   bool reached_goal_state;
   Notification notified;
   NotifyWhenServablesReachState(
@@ -212,7 +212,7 @@ bool ServableStateMonitor::WaitUntilServablesReachState(
         notified.Notify();
       });
   if (timeout > 0) {
-    WaitUntilServablesReachState(&notified, timeout)
+    WaitForNotificationWithTimeout(&notified, timeout_in_us);
   } else {
     notified.WaitForNotification();
   }
--- a/tensorflow_serving/core/servable_state_monitor.cc
+++ b/tensorflow_serving/core/servable_state_monitor.cc
@@ -196,7 +196,8 @@ void ServableStateMonitor::Notify(const NotifyFn& notify_fn) {
 bool ServableStateMonitor::WaitUntilServablesReachState(
     const std::vector<ServableRequest>& servables,
     const ServableState::ManagerState goal_state,
-    std::map<ServableId, ServableState::ManagerState>* const states_reached) {
+    std::map<ServableId, ServableState::ManagerState>* const states_reached,
+    const int timeout) {
   bool reached_goal_state;
   Notification notified;
   NotifyWhenServablesReachState(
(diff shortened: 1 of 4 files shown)


tensorflow_serving/model_servers/server_core.h
```diff
@@ -117,6 +117,9 @@ class ServerCore : public Manager {
     // Time interval between file-system polls, in seconds.
     int32 file_system_poll_wait_seconds = 30;
 
+    // Timeout for waiting on a servale to reach state.
+    int64 servable_state_monitor_timeout = -1;
```
chrisolston: nits:
- name suffix: timeout_micros (for consistency with other parts of tensorflow-serving codebase)
- don't use -1 to denote "no timeout". instead make the type optional<uint64>

tensorflow_serving/core/servable_state_monitor.cc
```diff
@@ -196,7 +196,8 @@ void ServableStateMonitor::Notify(const NotifyFn& notify_fn) {
 bool ServableStateMonitor::WaitUntilServablesReachState(
     const std::vector<ServableRequest>& servables,
     const ServableState::ManagerState goal_state,
-    std::map<ServableId, ServableState::ManagerState>* const states_reached) {
+    std::map<ServableId, ServableState::ManagerState>* const states_reached,
+    const int64 timeout_in_us) {
```
chrisolston: Please move this argument before "states_reached". (Our convention is to place const args before mutable ones.)

tensorflow_serving/model_servers/server_core.h
```diff
@@ -117,6 +117,9 @@ class ServerCore : public Manager {
     // Time interval between file-system polls, in seconds.
     int32 file_system_poll_wait_seconds = 30;
 
+    // Timeout for waiting on a servale to reach state.
```
chrisolston: Let's make the comment:
// Timeout for waiting for the models to load when the server starts or changes config.
(in this layer of our system we use the term "models" not "servables"; and the timeout is only on start and config change)
... and name the variable:
startup_and_config_reload_model_load_timeout (or something more concise if you can think of something :)

tensorflow_serving/core/servable_state_monitor.cc
```diff
@@ -210,7 +211,11 @@ bool ServableStateMonitor::WaitUntilServablesReachState(
         reached_goal_state = incoming_reached_goal_state;
         notified.Notify();
       });
-  notified.WaitForNotification();
+  if (timeout > 0) {
+    WaitForNotificationWithTimeout(&notified, timeout_in_us);
```
chrisolston: Given that you are merely forwarding to an existing utility method, and that testing this would be nontrivial, I don't think a test case is needed.

tensorflow_serving/model_servers/server_core.h
chrisolston: Don't you want to add a flag to model_servers/main.cc so you can control this from the command line when running the model-server binary?

rajhans:
I signed it!

rajhans:
Right, I see that. Bummer. AFAICS writing the code to set the state Notification times out will not be straightforward. I can take a stab at it later but don't have the cycles immediately. 

Do you still consider this a blocker to adding a service that provides an endpoint to synchronously reload the models (as you mentioned here #422)? If not, I can send out a PR that does that (we are already using some of that service in production).

chrisolston:
I guess the timeout is not a blocker. Do be sure your reload RPC PR documents clearly that it doesn't obey any timeout and may hang indefinitely waiting for newly-configured model(s) to load.

rajhans:
Ok, thanks Chris. Will send a PR with AdminService with the necessary comments.

Issue #700, opened 2017-12-21
Publish python 3 package


jkinkead:
The [tensorflow-serving-api package](https://pypi.python.org/pypi/tensorflow-serving-api/1.0.0) is only published for python 2. It should also be published for python 3.

This makes following [the tutorial](https://www.tensorflow.org/serving/serving_basic) impossible using python 3 - and the tutorial makes no note of the python 2 requirement.

chrisolston:
Sorry, tf-serving only supports python 2. See https://github.com/tensorflow/serving/pull/685/files

jkinkead:
The API package is just a collection of autogenerated classes from proto files. It *already* supports python 3 as-is, it just needs to be published as such.

illagrenan:
I can confirm, that the package **supports** Python 3 (we're using it on Python 3.6 without any changes). Unfortunately, the package is tagged as Python2 only on PyPI, so it is not possible to simply run `pip install ...`.

At this point we have to do:

1. Download `.whl` file from PyPI
2. Unzip package
3. Copy&Paste `tensorflow_serving/*` to our Python3 project

jkinkead:
Adding the `--universal` flag to [this command](https://github.com/tensorflow/serving/blob/c1ec43508ee57a5d6269116aba82d2a16d383c8a/tensorflow_serving/tools/pip_package/build_pip_package.sh#L53) should be sufficient to get the package labeled as cross-version:

```bash
python setup.py bdist_wheel --universal
```

jkinkead:
I'm happy to submit a patch fixing the script, but I can't publish or verify on my own.

Aspirinkb:
@illagrenan I Copy&Paste tensorflow_serving/* to my Python3 project, but `No moudle named 'tensorflow_serving.session_bundle'`.

chrisolston:
Folks, the issue is that, at this time, we cannot sign up for the ongoing support of python 3. Please understand that although it may be "easy", for a variety of reasons the cost to us is not zero and we do have to limit our commitments. We will revisit this decision in the future.

jkinkead:
To clarify: The request is that you label your existing, universal (Python3-compatible) build artifact as such.

We're asking for a label change, not an implementation change. The only Python3 support we're getting is from the gRPC project, not from TensorFlow.

hufangjian:
hi ,when the python3 api will be published,thanks.

illagrenan:
[...]

PR #721, opened 2018-01-09
Clarify that the API works in Python 3.

jkinkead:
Also add a passive-aggressive link to the closed issue.

Slightly tongue-in-cheek, but issue #700 really should be reopened.

### DIFF
--- a/tensorflow_serving/g3doc/setup.md
+++ b/tensorflow_serving/g3doc/setup.md
@@ -72,7 +72,12 @@ the `tensorflow-serving-api` PIP package using:
 pip install tensorflow-serving-api
 ```
 
-Note: The TensorFlow Serving Python API only supports Python 2.7. It does not support Python 3.
+Note: The TensorFlow Serving Python API
+[is only published for Python 2](https://pypi.python.org/pypi/tensorflow-serving-api/1.0.0),
+but will work for Python 3 if you either build it yourself, or download the Python 2 version,
+unzip it, and copy it into your Python 3 path. There was
+[an issue tracking publishing the API as Python 3](https://github.com/tensorflow/serving/issues/700),
+but it was closed.
 
 ## Installing using apt-get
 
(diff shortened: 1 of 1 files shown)


tensorflow_serving/g3doc/setup.md
```diff
@@ -72,7 +72,12 @@ the `tensorflow-serving-api` PIP package using:
 pip install tensorflow-serving-api
 ```
 
-Note: The TensorFlow Serving Python API only supports Python 2.7. It does not support Python 3.
+Note: The TensorFlow Serving Python API
+[is only published for Python 2](https://pypi.python.org/pypi/tensorflow-serving-api/1.0.0),
```
kirilg: Let's link to the package, not a specific version. Just https://pypi.python.org/pypi/tensorflow-serving-api

tensorflow_serving/g3doc/setup.md
```diff
@@ -72,7 +72,12 @@ the `tensorflow-serving-api` PIP package using:
...
+but will work for Python 3 if you either build it yourself, or download the Python 2 version,
+unzip it, and copy it into your Python 3 path. There was
```
kirilg: I reopened the issue. Let's remove the last line "There was [an issue tracking publishing the API as Python 3](https://github.com/tensorflow/serving/issues/700), but it was closed."

jkinkead: Would you like a link to the issue, without the snark, or should I just leave it off?

kirilg: I'm ok with linking to the issue as a feature request if you'd like. e.g. add something like:
"There's a [feature request](https://github.com/tensorflow/serving/issues/700) to publish a Python 3 package as well".

jkinkead:
Thanks, comments addressed!

kirilg:
Thanks! Merging now

PR #728, opened 2018-01-17
add new example 'twitter sentiment analysis'

robi56:


chrisolston:
Thanks for putting together an example that might be useful to members of the community! Unfortunately we cannot incorporate it into the repository because we cannot commit to supporting it. It would be great if you could make it available in some form to folks who are interested e.g. blog about it and link to the code patch from your blog.

PR #753, opened 2018-02-02
Remove tensorflow and tf_models as submodules now that they're linked…

kirilg:
… as bazel rules.


chenpingbupt:
this will cause docker building failed

PR #756, opened 2018-02-05
Add possibility to specify grpc channel args

sorhus:
This change makes it possible to pass channel arguments to gRPC when starting the model server.

With this change we can get the behaviour we need when using an L4 load balancer in front of an auto scaling group in AWS.

### DIFF
--- a/tensorflow_serving/model_servers/main.cc
+++ b/tensorflow_serving/model_servers/main.cc
@@ -265,16 +265,21 @@ class PredictionServiceImpl final : public PredictionService::Service {
   bool use_saved_model_;
 };
 
+struct ChannelArgument {
+  string key;
+  string value;
+};
+
 // Parses a comma separated list of gRPC channel arguments into list of pairs.
-std::vector<std::pair<string, string>> parseChannelArgs(
+std::vector<ChannelArgument> parseChannelArgs(
     const string& channel_arguments_str) {
   const std::vector<string> channel_arguments =
         tensorflow::str_util::Split(channel_arguments_str, ",");
-  std::vector<std::pair<string, string>> result;
+  std::vector<ChannelArgument> result;
   for (const string& channel_argument : channel_arguments) {
     const std::vector<string> key_val =
         tensorflow::str_util::Split(channel_argument, "=");
-    result.push_back(std::make_pair(key_val[0], key_val[1]));
+    result.push_back({key_val[0], key_val[1]});
   }
   return result;
 }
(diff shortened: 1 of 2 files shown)


tensorflow_serving/model_servers/main.cc
```diff
@@ -265,7 +265,7 @@ class PredictionServiceImpl final : public PredictionService::Service {
 };
 
 void RunServer(int port, std::unique_ptr<ServerCore> core,
-               bool use_saved_model) {
+               bool use_saved_model, string channel_arguments_str) {
```
chrisolston: "const string&" please

tensorflow_serving/model_servers/main.cc
```diff
@@ -276,6 +276,17 @@ void RunServer(int port, std::unique_ptr<ServerCore> core,
   builder.RegisterService(&model_service);
   builder.RegisterService(&prediction_service);
   builder.SetMaxMessageSize(tensorflow::kint32max);
+  const std::vector<string> channel_arguments =
```
chrisolston: Could you move the flag parsing to a separate helper function please. Then have a single callsite to builder.AddChannelArgument().

tensorflow_serving/model_servers/main.cc
```diff
@@ -276,6 +276,17 @@ void RunServer(int port, std::unique_ptr<ServerCore> core,
...
+      tensorflow::str_util::Split(channel_arguments_str, ",");
+  for (const string& channel_argument : channel_arguments) {
+    const std::vector<string> key_val =
+        tensorflow::str_util::Split(channel_argument, "=");
+    try { // Attempt to add as int, otherwise assume string.
```
chrisolston: Sorry, we don't use exceptions (try/catch) in our code-base. Please rewrite it using one of the functions in tensorflow/core/lib/strings/numbers.h that returns a bool instead of throwing an exception.

tensorflow_serving/model_servers/tensorflow_model_server_test.py
```diff
@@ -122,7 +122,8 @@ def RunServer(self,
                 model_name,
                 model_path,
                 batching_parameters_file='',
-                wait_for_server_ready=True):
+                wait_for_server_ready=True,
```
chrisolston: nit: please swap the order of these args s.t. wait_for_server_ready remains the last arg (and update callsites as needed)

tensorflow_serving/model_servers/main.cc
sorhus: I'm happy to move the parsing, but I don't see how I can achieve a single callsite since it has two signatures: https://github.com/grpc/grpc/blob/master/include/grpc%2B%2B/impl/channel_argument_option.h

I'm not very experienced with cpp, thanks for your help!

chrisolston: Oh, I see. I missed that. Please add some comments explaining what you are doing, since it's somewhat unusual. I gather that you want to handle a variety of arguments, some of which gRPC expects to have int values and others that will have string values, and you are hoping that you can distinguish based on int parsability.

The ideal approach would be to query gRPC for a given flag to find out its expected type (int or string), but assuming that isn't possible then what you are doing seems sensible. Please do document.

tensorflow_serving/model_servers/main.cc
```diff
@@ -264,8 +265,22 @@ class PredictionServiceImpl final : public PredictionService::Service {
   bool use_saved_model_;
 };
 
+// Parses a comma separated list of gRPC channel arguments into list of pairs.
+std::vector<std::pair<string, string>> parseChannelArgs(
```
chrisolston: Please don't use std::pair. Define a struct instead.

tensorflow_serving/model_servers/main.cc
```diff
@@ -264,8 +265,28 @@ class PredictionServiceImpl final : public PredictionService::Service {
   bool use_saved_model_;
 };
 
+struct ChannelArgument {
```
chrisolston: Please add some brief documentation about what this is for.

chrisolston: Please name it GrpcChannelArgument so it's clear that it pertains to gRPC.

chrisolston:
Looks good now. Please sign the CLA.

sorhus:
Thanks for working through this with me.

CLA is signed now.

sorhus:
ping @googlebot

chrisolston:

PR #757, opened 2018-02-06
Updated GPU dockerfile to be compatable with cuda 9.0 that tensorflow…

Modest-as:
Updated GPU dockerfile to be compatable with cuda 9.0 that tensorflow is using.


Modest-as:
Signed a CLA

Modest-as:
Signed a CLA

PR #759, opened 2018-02-07
Updated GPU dockerfile to be compatable with cuda 9.0

Modest-as:
Updated GPU dockerfile to be compatable with cuda 9.0 that tensorflow is using.


akshit-sharma:
Hello, I tried running your file to get the tensorflow serving in docker. 
I gave me an error on step 20/23 : 
Step 20/23 : RUN tensorflow/tools/ci_build/builds/configured GPU
 ---> Running in 473b8dab9119
/bin/sh: 1: tensorflow/tools/ci_build/builds/configured: not found

Can you please confirm if its only me ?

Thanks

Modest-as:
Hi @akshit-sharma,
I just double checked this morning and you are right, that dockefile was indeed broken. The problem was that since I've submitted this pull request, developers at tf serving changed their repo file structure and the corresponding dockerfile quite heavily, thus `tensorflow/tools/ci_build/builds/configured GPU` could not be found. I've merged my branch with the upstream one so if you take my dockefile now, you should be able to build no problem.

Cheers,

Modestas

netfs:
Please find updated docker files here:
https://hub.docker.com/r/tensorflow/serving/tags/

More info:
https://hub.docker.com/r/tensorflow/serving/

PR #766, opened 2018-02-12
add meta model proto files to pip package

yupbank:
Right now pip packages don't include meta model query `pb2.py` so that user can't query model status through grpc

This pr add them


nfiedel:

PR #774, opened 2018-02-19
Runtime model config reload gRPC API

Kindrat:
Added gRPC API for model config reload.

### DIFF
--- a/tensorflow_serving/apis/BUILD
+++ b/tensorflow_serving/apis/BUILD
@@ -157,6 +157,18 @@ serving_go_grpc_library(
 )
 
 serving_proto_library(
+    name = "model_management_proto",
+    srcs = ["model_management.proto"],
+    cc_api_version = 2,
+    go_api_version = 2,
+    java_api_version = 2,
+    deps = [
+        "//tensorflow_serving/config:model_server_config_proto",
+        "//tensorflow_serving/util:status_proto",
+    ],
+)
+
+serving_proto_library(
     name = "get_model_status_proto",
     srcs = ["get_model_status.proto"],
     cc_api_version = 2,
@@ -188,6 +200,7 @@ serving_proto_library(
     java_api_version = 2,
     deps = [
         ":get_model_status_proto",
+        ":model_management_proto",
     ],
 )
 
--- a/tensorflow_serving/apis/model_management.proto
+++ b/tensorflow_serving/apis/model_management.proto
@@ -1,0 +1,15 @@
+syntax = "proto3";
+
+import "tensorflow_serving/config/model_server_config.proto";
+import "tensorflow_serving/util/status.proto";
+
+package tensorflow.serving;
+option cc_enable_arenas = true;
+
+message ReloadConfigRequest {
+    ModelServerConfig config = 1;
+}
+
+message ReloadConfigResponse {
+    StatusProto status = 1;
+}
--- a/tensorflow_serving/model_servers/model_service_impl.cc
+++ b/tensorflow_serving/model_servers/model_service_impl.cc
@@ -17,20 +17,54 @@ limitations under the License.
 
 #include "tensorflow_serving/model_servers/get_model_status_impl.h"
 #include "tensorflow_serving/model_servers/grpc_status_util.h"
+#include "tensorflow_serving/util/status_util.h"
 
 namespace tensorflow {
-namespace serving {
+  namespace serving {
 
-::grpc::Status ModelServiceImpl::GetModelStatus(
-    ::grpc::ServerContext* context, const GetModelStatusRequest* request,
-    GetModelStatusResponse* response) {
-  const ::grpc::Status status = tensorflow::serving::ToGRPCStatus(
-      GetModelStatusImpl::GetModelStatus(core_, *request, response));
-  if (!status.ok()) {
-    VLOG(1) << "GetModelStatus failed: " << status.error_message();
-  }
-  return status;
-}
+    ::grpc::Status ModelServiceImpl::GetModelStatus(::grpc::ServerContext *context,
+                                                    const GetModelStatusRequest *request,
+                                                    GetModelStatusResponse *response) {
+      const ::grpc::Status status = tensorflow::serving::ToGRPCStatus(
+          GetModelStatusImpl::GetModelStatus(core_, *request, response));
+      if (!status.ok()) {
+        VLOG(1) << "GetModelStatus failed: " << status.error_message();
+      }
+      return status;
+    }
 
-}  // namespace serving
+    ::grpc::Status ModelServiceImpl::HandleReloadConfigRequest(::grpc::ServerContext *context,
+                                                               const ReloadConfigRequest *request,
+                                                               ReloadConfigResponse *response) {
+      ModelServerConfig server_config = request->config();
+      Status status;
+      switch (server_config.config_case()) {
+        case ModelServerConfig::kModelConfigList: {
+          const ModelConfigList list = server_config.model_config_list();
+
+          for (int index = 0; index < list.config_size(); index++) {
+            const ModelConfig config = list.config(index);
+            LOG(INFO) << "\nConfig entry"
+                      << "\n\tindex : " << index
+                      << "\n\tpath : " << config.base_path()
+                      << "\n\tname : " << config.name()
+                      << "\n\tplatform : " << config.model_platform();
+          }
+          status = core_->ReloadConfig(server_config);
+          break;
+        }
+        default:
+          status = errors::InvalidArgument("Invalid ServerModelConfig. Expecting ModelConfigList");
+      }
+
+      if (!status.ok()) {
+        LOG(ERROR) << "ReloadConfig failed: " << status.error_message();
+      }
+
+      const StatusProto status_proto = ToStatusProto(status);
+      *response->mutable_status() = status_proto;
+      return ToGRPCStatus(status);
+    }
+
+  }  // namespace serving
 }  // namespace tensorflow
--- a/tensorflow_serving/model_servers/model_service_impl.cc
+++ b/tensorflow_serving/model_servers/model_service_impl.cc
@@ -54,7 +54,7 @@ namespace tensorflow {
           break;
         }
         default:
-          status = errors::InvalidArgument("Invalid ServerModelConfig. Expecting ModelConfigList");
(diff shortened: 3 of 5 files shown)


tensorflow_serving/apis/model_management.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+import "tensorflow_serving/config/model_server_config.proto";
+import "tensorflow_serving/util/status.proto";
+
+package tensorflow.serving;
+option cc_enable_arenas = true;
+
+message LoadModelConfigurationRequest {
```
chrisolston: nit: let's call it ReloadConfigRequest for consistency with the ServerCore ReloadConfig() method name.

(and similarly ReloadConfigResponse below)

tensorflow_serving/apis/model_management.proto
```diff
@@ -0,0 +1,15 @@
...
+    ModelServerConfig configuration = 1;
```
chrisolston: nit: s/configuration/config for consistency with elsewhere in our codebase

tensorflow_serving/model_servers/main.cc
```diff
@@ -265,6 +272,42 @@ class PredictionServiceImpl final : public PredictionService::Service {
   bool use_saved_model_;
 };
 
+class ModelManagement final : public ModelManagementService::Service {
+  public:
+    explicit ModelManagement(ServerCore *core)
+            : core_(core) {}
+
+    grpc::Status Load(ServerContext *context,
```
chrisolston: nit: let's call this method HandleReloadConfigRequest()

tensorflow_serving/model_servers/main.cc
```diff
@@ -265,6 +272,42 @@ class PredictionServiceImpl final : public PredictionService::Service {
...
+                      const LoadModelConfigurationRequest *request,
+                      LoadModelConfigurationResponse *response) override {
+      const tensorflow::serving::ModelConfigList &list = request->configuration().model_config_list();
+      LOG(INFO) << "Reloading server config";
+
+      for (int index = 0; index < list.config_size(); index++) {
```
chrisolston: Let's drop this logging. Or move it inside ReloadConfig() (perhaps in a separate PR)

tensorflow_serving/model_servers/main.cc
```diff
@@ -265,6 +272,42 @@ class PredictionServiceImpl final : public PredictionService::Service {
...
 
+class ModelManagement final : public ModelManagementService::Service {
+  public:
+    explicit ModelManagement(ServerCore *core)
+            : core_(core) {}
+
+    grpc::Status Load(ServerContext *context,
+                      const LoadModelConfigurationRequest *request,
+                      LoadModelConfigurationResponse *response) override {
+      const tensorflow::serving::ModelConfigList &list = request->configuration().model_config_list();
```
chrisolston: Please use a switch statement to handle the various "oneof" cases, and throw an error if it's any case other than ModelConfigList.

tensorflow_serving/model_servers/main.cc
```diff
@@ -265,6 +272,42 @@ class PredictionServiceImpl final : public PredictionService::Service {
...
+      LOG(INFO) << "Reloading server config";
+
+      for (int index = 0; index < list.config_size(); index++) {
+        const tensorflow::serving::ModelConfig &config = list.config(index);
+          LOG(INFO) << "\nConfig entry"
+                  << "\n\tindex : " << index
+                  << "\n\tpath : " << config.base_path()
+                  << "\n\ttype : " << config.model_type()
```
chrisolston: model_type is deprecated so let's not include it in the logging

tensorflow_serving/model_servers/main.cc
```diff
@@ -265,6 +272,42 @@ class PredictionServiceImpl final : public PredictionService::Service {
...
+                  << "\n\tname : " << config.name()
+                  << "\n\tplatform : " << config.model_platform();
+      }
+      tensorflow::RunOptions run_options = tensorflow::RunOptions();
+      // By default, this is infinite which is the same default as RunOptions.
+      run_options.set_timeout_in_ms(DeadlineToTimeoutMillis(context->raw_deadline()));
+      const grpc::Status status =
+              tensorflow::serving::ToGRPCStatus(ModelManagementImpl::Load(core_, request, response));
+      if (!status.ok()) {
+        LOG(ERROR) << "Configuration Failed: " << status.error_message();
```
chrisolston: nit: ReloadConfig failed ....

tensorflow_serving/model_servers/model_management_impl.cc
```diff
@@ -0,0 +1,16 @@
+#include "tensorflow_serving/model_servers/model_management_impl.h"
+#include "tensorflow_serving/util/status_util.h"
+
+namespace tensorflow {
+    namespace serving {
+        Status ModelManagementImpl::Load(ServerCore *core,
+                                              const LoadModelConfigurationRequest* request,
+                                              LoadModelConfigurationResponse *response) {
+
+            const Status &status = core->ReloadConfig(request->configuration());
```
chrisolston: nit: drop the & in front off 'status' and 'statusProto'

tensorflow_serving/model_servers/model_management_impl.cc
```diff
@@ -0,0 +1,16 @@
...
+            const StatusProto &statusProto = ToStatusProto(status);
```
chrisolston: nit: please use the style status_proto instead of statusProto for consistency with the rest of our code-base (same in the rest of your PR, if you have any other camel-case variable names)

tensorflow_serving/model_servers/main.cc
```diff
@@ -265,6 +272,42 @@ class PredictionServiceImpl final : public PredictionService::Service {
   bool use_saved_model_;
 };
 
+class ModelManagement final : public ModelManagementService::Service {
```
chrisolston: Could this class be merged with ModelManagementImpl? It seems there is an unnecessary level of indirection.

vitalyli:
This would be great to have. Thanks. If you could review related issue: https://github.com/tensorflow/serving/issues/784

tensorflow_serving/model_servers/main.cc
Kindrat: I really don't know... If I remove separate service class, main.cc would be more complex, if move everything to separate service class I won't have access to DeadlineToTimeoutMillis method.

tensorflow_serving/model_servers/model_management_impl.cc
Kindrat: CLion autocompletion... Is it possible to create IDE-specific dir with project codestyle configs? Would save time for future reviews if most of the code would be using it automatically

tensorflow_serving/model_servers/main.cc
Kindrat: Do you mean gRPC anyof field type? The request is using single type. Or I'm just missing something

Kindrat:
Included all review requested changes except switch case.

tensorflow_serving/model_servers/main.cc
chrisolston: No. The ModelServerConfig proto has a one-of in it. The ModelConfigList is only one possible thing that might be in there (but the only one we need to handle at this time).

>>>
oneof config {
    ModelConfigList model_config_list = 1;
    google.protobuf.Any custom_model_config = 2;
  }
<<<

tensorflow_serving/apis/BUILD
```diff
@@ -156,6 +156,31 @@ serving_go_grpc_library(
...
+    go_api_version = 2,
+    java_api_version = 2,
+    deps = [
+        "//tensorflow_serving/config:model_server_config_proto",
+        "//tensorflow_serving/util:status_proto",
+    ],
+)
+
+serving_proto_library(
+    name = "model_management_service_proto",
```
chrisolston: Actually, rather than defining a new service, let's add this new RPC method to the existing ModelService service (defined in model_service.proto). The comment in that service would have to be updated to become broader (not just obtaining information, but also updating the set of models or other config).

tensorflow_serving/model_servers/main.cc
chrisolston: Leave as-is then.

Kindrat:
Moved handler to ModelService, added switch-case

tensorflow_serving/model_servers/model_service_impl.cc
```diff
@@ -17,20 +17,54 @@ limitations under the License.
...
-  if (!status.ok()) {
-    VLOG(1) << "GetModelStatus failed: " << status.error_message();
-  }
-  return status;
-}
-
-}  // namespace serving
+  namespace serving {
+
+    ::grpc::Status ModelServiceImpl::GetModelStatus(::grpc::ServerContext *context,
```
chrisolston: It would be better to leave the unchanged lines alone (not re-format them), so it's more clear which lines you actually changed in your PR. Thanks.

chrisolston:

vitalyli:
This is great to have. Would there be a corresponding update to tensorflow-serving-api python?

vitalyli:
Is this part of release 1.6 or 1.7? Also is there any test py client code? Trying to figure out how to add/remove models from model config.

Kindrat:
@vitalyli you should always pass full list of model sources, new entries would be deployed, missed entris - removed from serving. I guess, there is some major changes coming to deployment process according to discussion in https://github.com/tensorflow/serving/pull/809. Meanwhile, I've switched to my fork with add/remove APIs...

vitalyli:
@Kindrat thanks for your reply and implementing #774. not sure yet about #809. For a distributed model store system, I still need to push config changes into each serving instance. I already have a way to track which models are active and where model graphs are on the shared file system. until #774 there was not a clean way to deploy /undeploy new model name/type without restart, which is always problematic for  online serving. solution #774 seemed sufficient as I always know full set of models that are current. If you could add a little bit of documentation or python client apis example and which release this went into, that would be helpful. Like you mentioned release timelines are a bit unclear and what goes into each release also not clear. Thanks!

sparkingarthur:
Could you please give an example to show how to use it？

elchan:
Hi Folks, I was successful in using the python API to do the hot reload.
One thing I noticed is that 'model.conf' that I had originally on disk, does not get overwritten with the newly upload config.  Is this intentional or just a limitation?  Is there a way to cause the hot reload to overwrite the 'model.conf' that's on disk with the new one?

PR #782, opened 2018-02-23
Enable TensorFlow to build with Intel MKL for optimal performance on CPU

vamsi-sripathi:
TensorFlow uses Intel MKL to accelerate DNN primitives on CPU's. In the TensorFlow repository, to build with MKL support, it is a bazel config parameter, https://github.com/tensorflow/tensorflow/blob/master/tools/bazel.rc
We need the same config parameter, "build:mkl --define=using_mkl=true" to enabling TensorFlow serving with MKL


frallain:
+1 this tiny patch would be very useful.

vamsi-sripathi:
Since we don't need cuda to enable with MKL, I think it would be out of place to merge mkl build definitions with cuda.

vamsi-sripathi:
Ah, make sense. I made the changes.

gautamvasudevan:
Great! Can you resync and merge this PR? It has merge conflicts.

tools/bazel.rc
```diff
@@ -1,6 +1,8 @@
 build:cuda --crosstool_top=@org_tensorflow//third_party/gpus/crosstool
 build:cuda --define=using_cuda=true --define=using_cuda_nvcc=true
 
+build:mkl --define=using_mkl=true
+
```
gautamvasudevan: It would also make sense to do the same thing as TF does here: https://github.com/tensorflow/tensorflow/blob/master/tools/bazel.rc

tools/bazel.rc
```diff
@@ -1,6 +1,8 @@
 build:cuda --crosstool_top=@local_config_cuda//crosstool:toolchain
 build:cuda --define=using_cuda=true --define=using_cuda_nvcc=true
 
+build:mkl --define=using_mkl=true
+
```
gautamvasudevan: Thanks! Also noticed from: https://github.com/tensorflow/tensorflow/blob/master/tools/bazel.rc

# This config option is used to enable MKL-DNN open source library only,
# without depending on MKL binary version.
build:mkl_open_source_only --define=using_mkl_dnn_only=true

Seems like a good option here as well?

tools/bazel.rc
```diff
@@ -1,6 +1,12 @@
 build:cuda --crosstool_top=@local_config_cuda//crosstool:toolchain
 build:cuda --define=using_cuda=true --define=using_cuda_nvcc=true
 
+# Please note that MKL on MacOS or windows is still not supported.
+# If you would like to use a local MKL instead of downloading, please set the
+# environment variable "TF_MKL_ROOT" every time before build.
+build:mkl --define=using_mkl=true
+build:mkl -c opt
+
```
gautamvasudevan: Thoughts on match this part too?

# This config option is used to enable MKL-DNN open source library only,
# without depending on MKL binary version.
build:mkl_open_source_only --define=using_mkl_dnn_only=true

PR #788, opened 2018-03-01
Add weighted model selection

mrkm4ntr:
I added weighted model selection as a new servable.
I think this is helpful for A/B testing. 

## Example usage

1. Specify model weights as follows in `/tmp/selector/1/selector.conf`.
```
candidates: {
  name: "mnist1"
  weight: 0.1
}
candidates: {
  name: "mnist2"
  weight: 0.9
}
signatureGroups: {
  key: "predict_images"
  value: {
    signatures: {
      key: "mnist1"
      value: "predict_images"
    }
    signatures: {
      key: "mnist2"
      value: "predict_images"
    }
  }
}
```

2. Specify model config as follows in `models.conf`.
```
model_config_list: {
  config: {
    name: "mnist1",
    base_path: "/tmp/mnist_model1/",
    model_platform: "tensorflow",
  },
  config: {
    name: "mnist2",
    base_path: "/tmp/mnist_model2/",
    model_platform: "tensorflow",
  },
  config: {
    name: "weighted_mnist",
    base_path: "/tmp/selector",
    model_platform: "model_selector",
  },
}
```

3. Send a request using following model_spec.
```
name: weighted_mnist
signature_def: predict_images
```

## Note
Currently, I added this feature to only Predict I/F. If this is worth to merge, I will add this to another I/Fs and add tests.

### DIFF
--- a/tensorflow_serving/core/servable_handle.h
+++ b/tensorflow_serving/core/servable_handle.h
@@ -81,6 +81,12 @@ class ServableHandle {
   /// ServableHandle is null by default.
   ServableHandle() = default;
 
+  explicit ServableHandle(std::unique_ptr<UntypedServableHandle> untyped_handle)
+      : untyped_handle_(std::move(untyped_handle)),
+        servable_(untyped_handle_ == nullptr
+                  ? nullptr
+                  : untyped_handle_->servable().get<T>()) {}
+
   const ServableId& id() const { return untyped_handle_->id(); }
 
   // Smart pointer operations.
@@ -98,12 +104,6 @@ class ServableHandle {
 
  private:
   friend class Manager;
-
-  explicit ServableHandle(std::unique_ptr<UntypedServableHandle> untyped_handle)
-      : untyped_handle_(std::move(untyped_handle)),
-        servable_(untyped_handle_ == nullptr
-                      ? nullptr
-                      : untyped_handle_->servable().get<T>()) {}
 
   std::unique_ptr<UntypedServableHandle> untyped_handle_;
   T* servable_ = nullptr;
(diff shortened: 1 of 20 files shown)


chrisolston:
Hi, This is a pretty major proposed change. Before going further I'd like to understand the use-case. In particular, my leaning is that this sort of A/B testing support is better handled on the client side, while keeping the server simple (a given RPC request asks for model A, and gets model A). Is there a reason doing it on the client side is impractical?

mrkm4ntr:
@chrisolston Thanks for your reply.

The reason why we want to realize A / B test on the server side is as follows.
* Since ML related server-side team and a client-side team are clearly separated in my organization, we want to accelerate the verification cycle by realizing A / B test on the server side.
* Because there are many client-side implementation languages ​and platforms, the cost of constructing dynamic A / B test mechanisms on each is high.
* Since it is an environment where low latency is strongly demanded, we would like to complete the A / B test with serving alone without putting another component in front.

I think it's useful to have an option that A / B testing on the server side.

chrisolston:
I see. Thanks for explaining your rationale.

I looked a bit more at the change. It modifies quite a few layers of our system, including the RPC handlers, servable handle, and also introduces a new file format. Unfortunately I don't see it being worth it to incorporate this complexity into our code-base for a use-case that can be handled on the client side. We have all kinds of A/B testing happening at Google on the client side quite happily. I understand that other organizations may work differently, and that multi-language could be an issue, but I don't believe that justifies the large complexity I see in this change. I am not aware of any other users requesting this feature.

Thanks for your understanding. Perhaps if you can come up with a more surgical way to do this (strictly via the SourceAdapter registry mechanism and living under the tf::Session API, without all the other invasive changes), you could structure it as a library plug-in that you maintain in your own repo and link into TF-Serving.

PR #791, opened 2018-03-01
Add instructions for building SSL-enabled TensorFlow Serving.

jkinkead:
Note that the `bind()` call is a no-op. If I understand how this is working now, it's overriding a value inherited [from TensorFlow](https://github.com/tensorflow/tensorflow/blob/0abc4c9ecae912676f6070ca4b76b35c80351c26/tensorflow/workspace.bzl#L772-L777).

Partial fix for #193 .


AswathKiruba:
@jkinkead Thanks for your contribution. Could you please let me know the TLS protocol version (i.e. 1.1 to 1.3)  used in Tensorflow serving

netfs:
is this PR still relevant given #1011 is in master (and 1.10.0 release)? if not, can you please close this?

thanks!

jkinkead:
#1011 doesn't impact this change (although it adds much-needed configuration in support of it).

It might be that other work (perhaps in TensorFlow itself) makes this patch irrelevant. I'll see if I can confirm.

unclepeddy:
Closing due to lack of activity.

Issue #462, opened 2017-06-02
Export server metrics


FNegrello:
Hello,
Is there any way to have metrics (response time, prediction time, number of requests...) on the server ?
Thank you.

chrisolston:
Currently we don't have any built-in solution for that. Contributions welcome :)

Perhaps a good way to structure it would be to subclass ServableStateMonitor with something that also maintains & exports statistics.

ynqa:
How is this issue (or similar issue: #711) going?

ynqa:
@FNegrello @chrisolston 
How is this issue (or similar issue: #711) going?

ynqa:
If nobody push commits here now, I'd like to contribute.

Then, I also ask you the metrics definitions:
1. What are the elements output? (e.g. model_name, model_version, and elapsed_time per prediction...)
1. Metrics format (e.g. json, csv...)
1. How does it output the metrics?
    - @FNegrello case: Output metrics per calling `Predict` service, and use `LOG` (for stdout) or syslog in [this](https://github.com/Cdiscount/serving/commit/cfaae42556a81f8a2bbf98916e77b48332a25421) commit.
    - Another suggestion: Add the service for metrics to pull them from the monitoring systems.

cfregly:
the Prometheus metrics format - and pull model - is a good place to start.

works well with kubernetes, has c++ bindings, has a strong community, solid (turing-complete) query language, and great integration with a newly-updated Grafana UI.

for now, i would start with the following:

model_name

model_version

any custom labels/tags the user passes in

request counts

error counts (for each type of error you see in the TF Serving code)

graph execution time (aka prediction time)

in the future, i’d like to see the following:

custom ops for timing specific parts of the graph (similar in concept to the batch/unbatch ops, except start_timer/end_timer)

individual request tracing like ZipKin or Jaeger

model ensemble statistics (which models were involved in the final prediction)

a/b test metrics

ynqa:
> the Prometheus metrics format - and pull model - is a good place to start.

Sure. In this case, which is the better?
1. Add the service [here]
(https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/prediction_service.proto) for output metrics. but I'm not sure to pull metrics from gRPC service directly...
1. Preparing endpoint (e.g. `/metrics`) for HTTP requests, such as: https://github.com/grpc-ecosystem/go-grpc-prometheus

ynqa:
[...]

PR #800, opened 2018-03-12
Collect servable metrics

ynqa:
Related issue: #462

I implemented to manage servable metrics.
- Add a monitor to manage metrics using state event.
- Add gRPC service to fetch these metrics.

The client side example is [here](https://github.com/ynqa/serving/blob/metricsclient/tensorflow_serving/example/get_metrics_client.cc). It's also okay to include these code in this PR.


ynqa:
@googlebot I signed it!

ynqa:
I'll add tests after your reviewing once.

vinuraja:
Thanks @ynqa for the patch; appreciate it. We would like to support your use-case, but we aren't sure if this is the best way to do this. Please give us some time while we have some design discussions around this within the team. We will get back as soon as we have something to report.

ynqa:
@vinuraja Thanks. I wait for replies. And also please tell me if there is anything I should do.

chrisolston:
Hi, First of all tanks for the PR. I wanted to circle back now that we've had a chance to conduct some design discussions on our side. Our team is planning to implement this feature, but it will take some time to do it in a way that meets the various functionality and compatibility requirements from various stakeholders.  Please stay tuned.

PR #806, opened 2018-03-14
Fix a link to apt-get installation page

lc0:
Fix a link to apt-get installation page

### DIFF
--- a/tensorflow_serving/g3doc/serving_inception.md
+++ b/tensorflow_serving/g3doc/serving_inception.md
@@ -51,7 +51,7 @@ root@c97d8e820ced:/serving# bazel build -c opt tensorflow_serving/example/...
 ```
 
 Next we can either install a TensorFlow ModelServer with apt-get using the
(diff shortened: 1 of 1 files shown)


gautamvasudevan:
Not sure why this change is made - the original link (without some items switched to underscores) works for me and still seems correct:
https://github.com/tensorflow/serving/blob/master/tensorflow_serving/g3doc/setup.md#installing-using-apt-get

And the code that generates these on GitHub (https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb) doesn't appear to do any conversion to underscores.

lc0:
@gautamvasudevan You are right, on github it's fine, but the rendered version on tensorflow.org does not work - https://www.tensorflow.org/serving/setup#installing-using-apt-get

and the suggested one - https://www.tensorflow.org/serving/setup#installing_using_apt-get

lc0:
not sure, if it was already corrected in sources, but not yet on live website. What is the release cycle for the website docs? cc @gautamvasudevan

PR #797, opened 2018-03-06
List all model names

yupbank:
Add a grpc service to list all the model names available.

addressing https://github.com/tensorflow/serving/issues/795

Issue #671, opened 2017-11-27
service ping


mazorigal:
Hi,

is there any option to ping tensorflow serving, in order to check if the service is up and running ?
That is especially helpful in environments where each service monitored via ping and health endpoints.

Thanks,

bknl:
As far as I know the tensorflow_model_server does not support a health check RPC as in 

 https://github.com/grpc/grpc/blob/master/doc/health-checking.md

What I've used instead in my project are GetModelMetadataRequest calls to the servers to check for liveness and model presence.

mazorigal:
Thanks, I eventually ended up doing same, but still, feels as not the correct approach :-)
I think that they can implement sort of ping/health end point

kkozmic:
Another scenario (which I'm facing) is using TF Serving behind ELB (Network Load Balancer to be precise) where you don't control what the health request looks like.

Would there be any interest in adding that feature?

chrisolston:
Sounds like a reasonable feature to add. We would want to do it in a separate RPC service (not the model inference one). Contributions welcome :)

marceloboeira:
It does make a lot of sense, on my scenario, running under a k8s cluster, the `/healthz` endpoint helps a lot to keep the infrastructure monitored, rather than collecting errors from the clients.

ghisvail:
I'd be interested to know how `serving` is used in production without some form of health checking. What do people rely on in production then? I like @bknl's approach as a temporary fallback.

vitalyli:
If you know the list of model names, then it can be done using GetModelMetadata, except there is a bug in there: https://github.com/tensorflow/serving/issues/784

Zhiqiang-Ye:
I have the same requirement to get model status. I find some new protos in tf-serving 1.5 such as get_model_status.proto and model_service.proto, but have no idea how to use them. Reading the java code generated from these proto files, it does look like a separate RPC service mentioned by @chrisolston . I ask the same question on stackoverflow: [how-to-get-model-status](https://stackoverflow.com/questions/49051968/how-to-get-model-status-in-tensorflow-serving). Could anyone explain it a bit?

AswathKiruba:
@bknl @mazorigal  Could please let me know how to use GetModelMetadataRequest calls to the servers to check for liveness and model presence, as I am not able to use the GetModelStatus API #961 . Thank in advance

gmiejski:
[...]

PR #809, opened 2018-03-15
Added model config retrieve gRPC API

Kindrat:
In addition to config reload we now need to be able retrieve actual config from server - to modify it and send back when we don't wan't to unload some existing deployed model.

### DIFF
--- a/tensorflow_serving/apis/model_management.proto
+++ b/tensorflow_serving/apis/model_management.proto
@@ -13,3 +13,10 @@ message ReloadConfigRequest {
 message ReloadConfigResponse {
     StatusProto status = 1;
 }
+
+message GetConfigRequest {
+}
+
+message GetConfigResponse {
+    ModelServerConfig config = 1;
+}
--- a/tensorflow_serving/apis/model_service.proto
+++ b/tensorflow_serving/apis/model_service.proto
@@ -19,4 +19,7 @@ service ModelService {
   // Reloads the set of served models. The new config supersedes the old one, so if a
   // model is omitted from the new config it will be unloaded and no longer served.
   rpc HandleReloadConfigRequest(ReloadConfigRequest) returns (ReloadConfigResponse);
+
+  // Provides actual server config to use as source for further modifications via reload
+  rpc HandleGetConfigRequest(GetConfigRequest) returns (GetConfigResponse);
 }
--- a/tensorflow_serving/model_servers/model_service_impl.cc
+++ b/tensorflow_serving/model_servers/model_service_impl.cc
@@ -66,5 +66,12 @@ namespace tensorflow {
       return ToGRPCStatus(status);
     }
 
+    ::grpc::Status ModelServiceImpl::HandleGetConfigRequest(::grpc::ServerContext *context,
+                                                            const GetConfigRequest *request,
+                                                            GetConfigResponse *response) {
+      ModelServerConfig config = core_->GetConfig();
+      *response->mutable_config() = config;
+      return ToGRPCStatus(Status::OK());
+    }
   }  // namespace serving
 }  // namespace tensorflow
--- a/tensorflow_serving/model_servers/model_service_impl.h
+++ b/tensorflow_serving/model_servers/model_service_impl.h
@@ -22,6 +22,7 @@ limitations under the License.
 #include "tensorflow_serving/apis/model_service.pb.h"
 #include "tensorflow_serving/model_servers/server_core.h"
 #include "tensorflow_serving/apis/model_management.pb.h"
+#include "tensorflow_serving/apis/model.pb.h"
 
 namespace tensorflow {
   namespace serving {
@@ -37,6 +38,10 @@ namespace tensorflow {
       ::grpc::Status HandleReloadConfigRequest(::grpc::ServerContext *context,
                                                const ReloadConfigRequest *request,
                                                ReloadConfigResponse *response);
+
+      ::grpc::Status HandleGetConfigRequest(::grpc::ServerContext *context,
+                                            const GetConfigRequest *request,
+                                            GetConfigResponse *response);
 
     private:
       ServerCore *core_;
--- a/tensorflow_serving/model_servers/server_core.cc
+++ b/tensorflow_serving/model_servers/server_core.cc
@@ -471,6 +471,14 @@ Status ServerCore::ReloadConfig(const ModelServerConfig& new_config) {
   return Status::OK();
 }
 
+ModelServerConfig ServerCore::GetConfig() {
+  mutex_lock l(config_mu_);
+  ModelServerConfig config;
+  config = config_.operator=(config_);
+  VLOG(1) << "Sending server config: " << config.SerializeAsString();
+  return config;
+}
+
 Status ServerCore::CreateAdapter(
     const string& model_platform,
     std::unique_ptr<StoragePathSourceAdapter>* adapter) const {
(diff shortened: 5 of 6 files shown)


christisg:
Could you please describe specific use-cases for this?

Thanks

Kindrat:
@christisg for now config reload in serving acts as a full replacement of all that is currently deployed there. So to be able add single model in runtime you should get actual list from serving instance, add your new entry and pass it back for serving to pick new model. Same with removing deployed models

Kindrat:
Also it could be a replacement for https://github.com/tensorflow/serving/pull/797 as it provides such info

chrisolston:
@Kindrat Thanks for describing the use-case. I'm concerned that there could be races if you have multiple clients doing non-transactional read-modify-write operations on the set of configured models. I think a better approach would be to add AddModel()/DeleteModel() operations to the RPC interface as described in https://github.com/tensorflow/serving/issues/811

Kindrat:
@chrisolston I disagree. Since serving is an application server, it should provide API for managing models. Retrieve endpoint could be also used for checking if model of specified name and version is already deployed. Or to get absolute oath of model directory to cleanup resources after undeploying model from serving instance.

chrisolston:
I want to keep the surface area of this emerging model control API minimal. Would the ListModels proposal in https://github.com/tensorflow/serving/pull/797 meet your needs?

Kindrat:
Well, It does not provide model versions and base path. Without version you can't say if required one is deployed and without path you can't cleanup resources. I agree, that API should be simple, yet sufficient.

Mb, it's preferrable to delegate cleanup to serving itself - e.g. remove all related files after model is undeployed by API call? But it's still not enough to return only model names.

chrisolston:
After a discussion with other TF-Serving team members about this proposal and several others in the same space around config reading/updating, we have the following observation and proposal:

Observation: Using a ModelServer process as a state store (reading which models are configured, adding/removing individual models from the configured set, and so on) is brittle because the state is not persisted -- and you would lose the state upon a crash.

Proposal: Perhaps a more robust approach would be a convention in which the config resides in a persistent storage medium (e.g. google cloud storage, or any file system), and the ModelServer polls to discover config changes. The ground truth about the config would be in the storage system, not the ModelServer job. And for making delta config changes (add/remove model and so on) you would (atomically) update the file -- possibly via a separate service with an AddModel()/RemoveModel() RPC interface.

So, we are going to push back on proposals to add RPCs to read and delta-update the model config, but in lieu of those we would be open to a PR that adds a thread that polls a config file and calls ReloadConfig() with the latest file content.

yupbank:
Yet we can use some helper to get the state of ModelServer?  That's why i had a pr just to just get the model names in ModelServer memory. We are adding a http layer on top of TF-Serving, which can provide a more nice interface, REST API, API Specs for all the models available to server.   https://github.com/kubeflow/kubeflow/tree/master/components/k8s-model-server/http-proxy

Kindrat:
Do you mean gRPC service in serving?

chrisolston:
No, it would be a separate service to facilitate and synchronize config changes.

Kindrat:
So, you require deploying two applications to use serving? Just to change the config? That's really strange. Why not to include persistent layer client in serving and let it hold this service, so that instance would know about update request immidiatelly? Instead of polling from some remote source, updated in another client. 

By adding new implementations of persistent layer, it would be possible to configure serving use any of available storages - e.g. S3 and other.

chrisolston:
Handling config management in a separate system is not an uncommon pattern. See e.g. https://sysgears.com/articles/managing-configuration-of-distributed-system-with-apache-zookeeper/.

I can imagine packaging the config management system together with the TF-Serving binary, as you say, if that makes things more convenient for some users, at the expense of not allowing the config management layer to be generally used across multiple systems being configured. But that coupling would complicate the TF-Serving binary unnecessarily.

It seems to me that if you move away from a static config and are dynamically changing the models being served, then you must be in a somewhat complex environment that has multiple systems/layers anyway -- you aren't going to have everything in one binary as it is. Sorry if I'm missing something -- it can be hard to talk in general terms since we can't see eachothers' architectures :)

Kindrat:
That's a great example, unless the great production-ready app Apache Kafka is also hiding zookeeper behind own API. It's great to plan fully distribute system with load balancing, external configuration (I prefer https://github.com/Netflix/archaius in this case), but serving does not provide even basic API at the moment. I mean health checks, metrics, detailed logging (not the graph computations in VLOG - rather structured by log level messages).

I'd rather focus on basic APIs and replace these in-memory updates with integration with some external storage lately.

My biggest concern at the moment is that client would need two separate connections to separate services - it means we need two ports on firewall, load balancer, proxy, and more resources on client side. We have two use cases: one with java app (and it's OK to have as many connections as we need), but the other is `C` application with very small footprint and I would try to avoid increasing resource requirements for it's runtime as much as possible. Even if it would mean completely focus on supporting our fork of serving...

nfiedel:
Hi @Kindrat , @chrisolston , All, 
  First, apologies for not having a better end-to-end story for model management (e.g. APIs, implementation, documentation). This is something that some of our largest users and contributors have wanted, and we are interested in better supporting this area.

  A key challenge we face is how to stay focused to avoid scope-creep, while also enabling robust production usage. For this area in particular, it seems that there is a requirement that configs be persisted in a storage system outside of TF Serving. This is needed to robustly handle task/vm/machine failures and autoscaling (bringing up new tasks). It is also a best practice to be able to backup or restore a serving config, for example when cloning an environment to run a canary or staging environment, or simply as a disaster-recovery mechanism. 

  With the common understanding that configs are persisted in a common file store (S3, GCS, ...), the architecture question is where to orchestrate updating of configs. Very open to learning more about your use-cases, architecture (i.e. what client system/programming language will be making the updates), and ACLs (i.e. do you have multiple clients, and if so which of them are allowed to perform mutations).
  Thanks!
  Noah

Kindrat:
External config service is OK for decentrilized system, but I'm afraid of moving towards wrong direction at this point. While the API is still very basic and does not provide even minimum features for remote management, working on extracting config microservice seems like priority issue. Serving is even not synchronised with TF releases. Mb it would be much more easier to discuss such updates if we could see project roadmap at least...

And again - about extracting microservices in serving - it's great but it would even greater to have separate API layer in front of all those servises b/c it makes deployment terrible - you need to watch for updates of several APIs of several services and update several clients while the whole thing could be similar to Kafka API, where you do not need to worry about details of broker cluster management on client side. I guess, we can do it and contribute later.

So as for use cases: there is a cloud app allowing running data analytics requests. One of processing nodes provides TF-related processing (runnig over pre-defined pre-trained models). Those nodes are created on-demand and may hold some amount of running models at same time. How many models and what exactly (types, versions) - is defined by user request. So we need to be able deploy and cleanup models on serving instances dynamically in runtime.

I would really like to contribute everything we will add to serving, but I'm confused about project development process. There is no discussion in user-created issues, no labels on issues, no release road map and no synchronisation with tensorflow releases. Even now it looks like model management API would be postponed for unknown period of time in favour of implementing external config... Why not https://github.com/tensorflow/serving/issues/671 ?

vitalyli:
For my use-case, PR #774 answers model configuration. Next to each instance I have my own asynchronous controller, which knows, or has a way to find (think of distributed database or equivalent setup), exactly which models are active and where the model graphs are on the shared FS. I'm not yet able to run it however, for some reason python pip is not installing tensorflow-serving-api properly, but logically reconfiguring full set of models is fine with one caveat, that existing models do not get interrupted while new entries are added as existing ones may be serving online traffic.

vitalyli:
@nfiedel is this closed as not to be done or there is a plan to do it differently?

Kindrat:
@nfiedel any news on this feature?

Issue #804, opened 2018-03-14
create pip package path issue


brentian:
I'm trying to build the pip package from source using bazel

compiling is fine using the guide:

```sh
bazel test -c opt tensorflow_serving/...
```
then I tried to create pip package

```sh
bazel-bin/tensorflow_serving/tools/pip_package/build_pip_package /tmp/
```
with error
```
cp: cannot stat 'bazel-serving/tensorflow_serving/apis/*_pb2.py': No such file or directory
```
 I looked into the script at serving/tensorflow_serving/tools/pip_package/build_pip_package.sh and found the problem is that it uses path `bazel-serving` instead of `bazel-tensorflow-serving`.

I changed output directory `bazel-tensorflow-serving` back to `bazel-serving`  and it works well. I suppose it could also be a quick fix by changing paths in the script.

gautamvasudevan:
The script assumes you didn't clone into the directory you're in. The default directory that will be created is ./serving/, as this script is called from another script that follows that pattern.

PR #810, opened 2018-03-15
change build_pip_package.sh; fix path

brentian:
issue #804


brentian:
I signed it! @googlebot

gautamvasudevan:
build_pip_package is called from other scripts that typically pull not to the local directory. A default git pull will create a /serving directory, which is why this is the way it is. Thanks though!

PR #849, opened 2018-04-14
Sync to TF r1.7

vinuraja:


Hijiao:
hello,@vinuraja I think you missed something while syn to TF r1.7 ,  look at #863

PR #863, opened 2018-04-27
use bazel 0.10.0 in Dockerfile

Hijiao:
The Dockerfile currently uses bazel 0.5.4, but  for TensorFlow 1.7.0, the minimum version of bazel required is 0.10.0.  I update the version everywhere following  #620


Yakings:
Good, you save my day!

junjin8433:
I also encountered the same problem in my projects. This really puzzled me for a while and thank you so much.

PR #885, opened 2018-05-18
Support runtime model_config_file reload

pockyym:
Support runtime model_config_file reload.
when using --model_config_file, tf-serving starts a new thread that polls filesystem and triggers loading or unloading action by calling ServerCore::ReloadConfig(). It has been verified on our production prediction engin for easily releasing new models, makes it convenient for both multiple-model and multiple-version a/b test.

### DIFF
--- a/tensorflow_serving/model_servers/config_reloader.cc
+++ b/tensorflow_serving/model_servers/config_reloader.cc
@@ -1,0 +1,77 @@
+/* Copyright 2018 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "tensorflow_serving/model_servers/config_reloader.h"
+
+#include <string>
+#include <utility>
+#include <thread>
+#include <sys/stat.h>
+
+#include "tensorflow/core/lib/core/status.h"
+#include "tensorflow/core/platform/macros.h"
+#include "tensorflow/core/platform/mutex.h"
+#include "tensorflow/core/platform/types.h"
+#include "tensorflow_serving/config/model_server_config.pb.h"
+#include "tensorflow_serving/core/source.h"
+#include "tensorflow_serving/core/storage_path.h"
+#include "tensorflow_serving/sources/storage_path/file_system_storage_path_source.h"
+
+namespace tensorflow {
+namespace serving {
+
+bool ConfigReloader::Start() {
+  exit_ = false;
+  stop_loading_ = false;
+  loading_thread_.reset(new std::thread(std::bind(&ConfigReloader::Run, this)));
+  return true;
+}
+
+bool ConfigReloader::Stop() {
+  exit_ = true;
+  stop_loading_ = true;
+  return true;
+}
+
+void ConfigReloader::Run() {
+  while (!exit_) {
+    Reload();
+    // sleep
+    std::this_thread::sleep_for(std::chrono::seconds(kInterval));
+  }
+}
+
+void ConfigReloader::Reload() {
+  if (!Env::Default()->FileExists(file_path_).ok()) {
+    LOG(INFO) << "Config file not exist: " << file_path_;
+    return;
+  }
+  struct stat file_stat;
+  if (-1 == stat(file_path_.c_str(), &file_stat)) {
+    LOG(INFO) << "Failed to stat: " << file_path_;
+    return;
+  }
+  time_t modifiled_time = file_stat.st_mtime;
+  if (modifiled_time <= last_modified_time_) {
+    return;
+  }
+  last_modified_time_ = modifiled_time;
+  if (reload_func_) {
+    reload_func_();
+  }
+}
+
+}  // namespace serving
+}  // namespace tensorflow
--- a/tensorflow_serving/model_servers/config_reloader.h
+++ b/tensorflow_serving/model_servers/config_reloader.h
@@ -1,0 +1,70 @@
+/* Copyright 2018 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#ifndef TENSORFLOW_SERVING_MODEL_SERVERS_CONFIG_LOADER_H_
+#define TENSORFLOW_SERVING_MODEL_SERVERS_CONFIG_LOADER_H_
+
+#include <string>
+#include <utility>
+#include <thread>
+
+#include "tensorflow/core/lib/core/status.h"
+#include "tensorflow/core/platform/macros.h"
+#include "tensorflow/core/platform/mutex.h"
+#include "tensorflow/core/platform/types.h"
+#include "tensorflow_serving/config/model_server_config.pb.h"
+#include "tensorflow_serving/core/source.h"
+#include "tensorflow_serving/core/storage_path.h"
+#include "tensorflow_serving/sources/storage_path/file_system_storage_path_source.h"
+
+namespace tensorflow {
+namespace serving {
+
+class ConfigReloader {
+ public:
+  ConfigReloader(const std::string& file_path, bool stop_loading = false)
+      : file_path_(file_path), stop_loading_(stop_loading) {}
+  virtual ~ConfigReloader() {
+    if (loading_thread_) {
+      loading_thread_->join();
+    }
+  }
+  ConfigReloader(const ConfigReloader&) = delete;
+  ConfigReloader& operator=(const ConfigReloader&) = delete;
+
+  virtual bool Start();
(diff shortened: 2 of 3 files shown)


pockyym:
I signed it!

chrisolston:
Hi, We already have a mechanism for reloading the config -- it's via the RPC endpoint here [1]. Would that work for your use-case?

[1] https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/model_service.proto#L22

pockyym:
It seems like that client-side needs to monitor the config and send reload request to server-side. why not do it on the server-side?

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+
+  virtual bool Start();
+  virtual bool Stop();
+  const string& GetFilePath() { return file_path_; }
+  void SetReloadFunc(std::function<void()> reload_func) {
+    reload_func_ = reload_func;
+  }
+
+ protected:
+  const int32_t kInterval{ 1 };
```
chrisolston: Please make this configurable (vs. hard-coded), and set the default to 60 seconds (to avoid accidentally eating up a lot of FS resources (or even DoSing) in a deployment that has a large number of model server binaries running).

chrisolston:
Yeah, I can see the case for a server-side alternative. I'll review the PR.

tensorflow_serving/model_servers/main.cc
```diff
@@ -70,6 +70,7 @@ limitations under the License.
 #include "tensorflow_serving/apis/prediction_service.pb.h"
 #include "tensorflow_serving/config/model_server_config.pb.h"
 #include "tensorflow_serving/core/availability_preserving_policy.h"
+#include "tensorflow_serving/model_servers/config_reloader.h"
```
chrisolston: You're going to have to edit the BUILD file. I don't understand how it can work otherwise to add the new source files. Maybe you forgot to incorporate the BUILD file changes into the PR?

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+#include "tensorflow_serving/core/source.h"
+#include "tensorflow_serving/core/storage_path.h"
+#include "tensorflow_serving/sources/storage_path/file_system_storage_path_source.h"
+
+namespace tensorflow {
+namespace serving {
+
+class ConfigReloader {
+ public:
+  ConfigReloader(const std::string& file_path, bool stop_loading = false)
```
chrisolston: Please document what "stop_loading" does.

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+      : file_path_(file_path), stop_loading_(stop_loading) {}
```
chrisolston: nit: don't inline constructors. Please move to the .cc file.

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+      loading_thread_->join();
+    }
+  }
+  ConfigReloader(const ConfigReloader&) = delete;
+  ConfigReloader& operator=(const ConfigReloader&) = delete;
+
+  virtual bool Start();
+  virtual bool Stop();
+  const string& GetFilePath() { return file_path_; }
+  void SetReloadFunc(std::function<void()> reload_func) {
```
chrisolston: set_reload_func() please, for style consistency

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+    reload_func_ = reload_func;
+  }
+
+ protected:
```
chrisolston: Please use "private" instead of "protected", and eliminate the "virtual" keywords. We don't need to set this up for inheritance do we?

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+  const int32_t kInterval{ 1 };
+  volatile bool exit_;
```
chrisolston: Please don't use "volatile".

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+#include "tensorflow/core/platform/types.h"
+#include "tensorflow_serving/config/model_server_config.pb.h"
+#include "tensorflow_serving/core/source.h"
+#include "tensorflow_serving/core/storage_path.h"
+#include "tensorflow_serving/sources/storage_path/file_system_storage_path_source.h"
+
+namespace tensorflow {
+namespace serving {
+
+class ConfigReloader {
```
chrisolston: Please document what the class does.

// <explanation of this class>
class ConfigReloaded {
...

tensorflow_serving/model_servers/config_reloader.cc
```diff
@@ -0,0 +1,77 @@
+/* Copyright 2018 Google Inc. All Rights Reserved.
```
chrisolston: Please add a unit test for this class.

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+    if (loading_thread_) {
+      loading_thread_->join();
+    }
+  }
+  ConfigReloader(const ConfigReloader&) = delete;
+  ConfigReloader& operator=(const ConfigReloader&) = delete;
+
+  virtual bool Start();
+  virtual bool Stop();
+  const string& GetFilePath() { return file_path_; }
```
chrisolston: file_path() please, for style consistency

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+      : file_path_(file_path), stop_loading_(stop_loading) {}
+  virtual ~ConfigReloader() {
+    if (loading_thread_) {
+      loading_thread_->join();
+    }
+  }
+  ConfigReloader(const ConfigReloader&) = delete;
+  ConfigReloader& operator=(const ConfigReloader&) = delete;
+
+  virtual bool Start();
```
chrisolston: This API is easy to misuse (you have call ctor then SetReloadFunc() then Start() in that order). Instead, please structure the API as follows, s.t. Create() takes all the params and starts it, and the dtor stops it:

public:
// static factory method. does what the ctor and Start() currently do
static Status Create(file_path, reload_func, std::unique_ptr<ConfigReloader>* reloader);

dtor here, which does what Stop() currently does.

// (no Start() or Stop())

private:
  private ctor here; called by Create()

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+  virtual bool Stop();
+  const string& GetFilePath() { return file_path_; }
+  void SetReloadFunc(std::function<void()> reload_func) {
+    reload_func_ = reload_func;
+  }
+
+ protected:
+  const int32_t kInterval{ 1 };
+  volatile bool exit_;
+  volatile bool stop_loading_;
```
chrisolston: Please add a comment before each class variable to explain what it does.

chrisolston: Do you need all these? Could you eliminate start_loading_ and stop_loading_?

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+  std::unique_ptr<std::thread> loading_thread_;
+  std::string file_path_;
+  time_t last_modified_time_{ 0 };
```
chrisolston: nit: use =0 instead of {0}

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+  void SetReloadFunc(std::function<void()> reload_func) {
+    reload_func_ = reload_func;
+  }
+
+ protected:
+  const int32_t kInterval{ 1 };
+  volatile bool exit_;
+  volatile bool stop_loading_;
+  std::unique_ptr<std::thread> loading_thread_;
+  std::string file_path_;
```
chrisolston: nit: const

tensorflow_serving/model_servers/config_reloader.h
```diff
@@ -0,0 +1,70 @@
...
+  time_t last_modified_time_{ 0 };
+  std::function<void()> reload_func_;
+
+  virtual void Run();
```
chrisolston: style nit: please put the methods before the variables, like so:

private:
  private methods...
  private variables...

chrisolston: Please put a comment before every method to explain what it does.

tensorflow_serving/model_servers/main.cc
```diff
@@ -263,6 +282,7 @@ class PredictionServiceImpl final : public PredictionService::Service {
   ServerCore* core_;
   std::unique_ptr<TensorflowPredictor> predictor_;
   bool use_saved_model_;
+  std::unique_ptr<ConfigReloader> loader_;
```
chrisolston: nit: config_reloader_

tensorflow_serving/model_servers/main.cc
```diff
@@ -176,10 +178,27 @@ int DeadlineToTimeoutMillis(const gpr_timespec deadline) {
...
  public:
-  explicit PredictionServiceImpl(ServerCore* core, bool use_saved_model)
+  explicit PredictionServiceImpl(ServerCore* core,
+		                 bool use_saved_model,
+				 const string& model_config_file)
       : core_(core),
         predictor_(new TensorflowPredictor(use_saved_model)),
-        use_saved_model_(use_saved_model) {}
+        use_saved_model_(use_saved_model) {
+    if (!model_config_file.empty()) {
```
chrisolston: I'm pretty sure we'll want a flag to control whether to do this dynamic config reloading. We'll also want to be able to control the polling period. Maybe we can combine these into a single flag: config_reload_polling_period_seconds /* setting to 0 disables dynamic config reloading */. And have it default to 0 (no polling/reloading). WDYT?

gautamvasudevan:
Closing due to lack of activity.

marvinxu-free:
is there a python way to reload model update from base dir?

PR #991, opened 2018-07-16
Add support for Python 3 while still compatible with Python 2.7

liuqun:
Since Python 3 has dropped `xrange()`, we should always use `range()` in the future.


liuqun:
I signed it!

Issue #1057, opened 2018-08-15
Documents for runtime model config reload


sparkingarthur:
I have the same needs as issue #380 , which means that I need to be able to dynamically add new models for serving. Actually i find a solution in this issue ( which is add a thread to check directory and rebuild a ModelServerConfig and use the ServerCore::ReloadConfig method to reload the config. The problem is that my C++ is already a but rusty by now and I don't fully understand yet all parts of TF serving, including the BUILD file of bazel, so i try to find a simpler way to meet my needs. Then i find PR #774 , but i don't know how to use this api exactly. Could you please give me an example or some documents to show how to use this api to reload model config at runtime exactly? Thx a lot.

ref:[https://github.com/tensorflow/serving/pull/774](#774 )

gautamvasudevan:
Duplicate of #380.

Issue #964, opened 2018-06-28
How to specify the ModelConfigList for HandleReloadConfigRequest ?


chenjien:
Hi, folks.

There are two services in model_service.proto : GetModelStatus and HandleReloadConfigRequest.

I can find a example of call to GetModelStatus in tensorflow_model_server_test.py, is there any example of call to HandleReloadConfigRequest? 

I don't know how to specify the ModelConfigList in python.

Issue #380, opened 2017-03-28
How to use the custom_model_config to load models at runtime


Hello-Frankey:
The [tensorflow serving advanced tutorial](https://tensorflow.github.io/serving/serving_advanced) says the models to be loaded can be declared either through model_config_list or through dynamic_model_config as follows:
> ServerCore::Create() takes a ServerCore::Options parameter. Here are a few commonly used options:
ModelServerConfig that specifies models to be loaded. Models are declared either through model_config_list, which declares a static list of models, or through dynamic_model_config, which declares a dynamic list of models that may get updated at runtime.

Now, i can load multiple models through model_config_list with a text config file like this:

model_config_list: {
    config: {
        name: "mnist",
        base_path: "/tmp/monitored/inception_model",
        model_platform: "tensorflow"
    },
    config: {
        name: "inception",
        base_path: "/tmp/monitored/inception_model",
        model_platform: "tensorflow"
    }
}

But, how to use the dynamic_model_config to load some models at runtime? Any example or detail explanation about that? I also read the following comments in config/model_server_config.proto. It says the custom model config is fetched dynamically at run time through network RPC, custom service, etc. Okay, i am still confused...

> // ModelServer config.
message ModelServerConfig {
  **// ModelServer takes either a static file-based model config list or an Any
  // proto representing custom model config that is fetched dynamically at
  // runtime (through network RPC, custom service, etc.).**
  oneof config {
    ModelConfigList model_config_list = 1;
    google.protobuf.Any custom_model_config = 2;
  }
}

Hello-Frankey:
Still have no idea about how to use the dynamic_model_config, maybe it's still unsupported? I modified main.cc & server_core.cc in the tensorflow_serving/model_servers and implemented automatic loading function i want.

skonto:
Have the same issue, I dont see dynamic_model_config anywhere in the code, so probably it is not supported. Btw the only extension supported is to write plugins to manage the new versions:
https://tensorflow.github.io/serving/custom_source

kirilg:
Seems we had one tutorial that still had the old name of dynamic_model_config which was renamed to custom_model_config. I'll update the documentation.

A custom_model_config can be used for anything that you can't do with a standard model config that points to models on disk. A common case is if you have your own custom sources for example. It was originally called dynamic because you can configure a source that dynamically changes the list of models to load based on some external signals, but was renamed to something more general (custom_model_config) since it may or may not be dynamic.

[...]

PR #1065, opened 2018-08-21
add an example for hot reload using gRPC

sparkingarthur:
Many issues discuss how to update the models during runtime such as #380 #422.
And I found that a useful PR #774 is merged into master. However, many people are not familiar with protobuf (especially ModelConfig.proto, ModelServerConfig.proto, etc) and grpc such as #964 #1057  , and don't know how to make a gRPC hot-reload request. This is an example written in python, showing how to make a gRPC hot-reload request to reload the models.

### DIFF
--- a/tensorflow_serving/example/hot_reload_grpc_example.py
+++ b/tensorflow_serving/example/hot_reload_grpc_example.py
@@ -1,0 +1,38 @@
+#coding=utf-8
+from tensorflow_serving.apis import model_service_pb2
+from tensorflow_serving.apis import model_service_pb2_grpc
+from tensorflow_serving.apis import model_management_pb2
+from tensorflow_serving.config import model_server_config_pb2
+from tensorflow_serving.util import status_pb2
+
+import grpc
+
+def run():
+  channel = grpc.insecure_channel('yourip:yourport')
+  stub = model_service_pb2_grpc.ModelServiceStub(channel)
+  request = model_management_pb2.ReloadConfigRequest()  ##message ReloadConfigRequest
+  model_server_config = model_server_config_pb2.ModelServerConfig()
+
+  config_list = model_server_config_pb2.ModelConfigList()##message ModelConfigList
+  
+  one_config = config_list.config.add() #####try to add one model config
+  one_config.name= "test1"
+  one_config.base_path = "/home/model/model1"
+  one_config.model_platform="tensorflow"
+
+  model_server_config.model_config_list.CopyFrom(config_list) c
+
+  request.config.CopyFrom(model_server_config)
+
+  print(request.IsInitialized())
+  print(request.ListFields())
+
+  responese = stub.HandleReloadConfigRequest(request,10)
+  if responese.status.error_code == 0:
+      print("Reload sucessfully")
+  else:
+      print("Reload failed!")
+      print(responese.status.error_code)
+      print(responese.status.error_message)
+
+run()


sparkingarthur:
I signed it. Thx a lot for your work!

sparkingarthur:
ok, I'm glad to know that you will put out more examples coupled with documentation, thx a lot anyway.

Rojina99:
@sparkingarthur I am working on the same problem. I have already loaded the model with model with static model_config_list and now want to add new model dynamically as your example above.But the problem is , output of all of these lines come out empty:

```
request = model_management_pb2.ReloadConfigRequest()  ##message ReloadConfigRequest
model_server_config = model_server_config_pb2.ModelServerConfig()
config_list = model_server_config_pb2.ModelConfigList()
```
So the only new model I passed will be loaded replacing previous one. 

So I have few questions:
- From which version the serving supports this feature?
- Can you suggest me any reason why previous model are not loaded?

tensorflow_serving/example/hot_reload_grpc_example.py
```diff
@@ -0,0 +1,38 @@
...
+  model_server_config = model_server_config_pb2.ModelServerConfig()
+
+  config_list = model_server_config_pb2.ModelConfigList()##message ModelConfigList
+  
+  one_config = config_list.config.add() #####try to add one model config
+  one_config.name= "test1"
+  one_config.base_path = "/home/model/model1"
+  one_config.model_platform="tensorflow"
+
+  model_server_config.model_config_list.CopyFrom(config_list) c
```
Rojina99: c is not required here

yingjieyao:
@Rojina99 
The answer to the second question maybe
```
model_management_pb2.ReloadConfigRequest
```
Maybe we should add previous model as well as the new added model through config.add()?
Any better idea?

Rojina99:
@yingjieyao Yes previously I thought the ReloadConfigRequest might also give the configuration of models running at servers. But later looking further, I got to know that it doesnot carry configuration list. We have to pass configuration of previous model along with new one to load any new model to server.

tensorflow_serving/example/hot_reload_grpc_example.py
sparkingarthur: 1.it seems that the serving supports this feature from r1.7？
2.you should pass the complete model list as a model_config_list pb

you can find more information in my github: https://github.com/sparkingarthur/serving/tree/master/custom_scripts

elvys-zhang: How can I set model_version_policy to 'all' or 'specific' ?

karlschriek:
> I totally agree that additional code examples and documentation would be helpful. Thanks for taking the initiative on that front and sending a PR. We are working on our side to put out additional code examples coupled with documentation, so please stay tuned for that.

Just popping in to say that I had to search for ages to find an example of how to do this and stumbled across this PR by chance (which finally solved my problem). I'm wondering why you closed the pull request? Do you still intend to release some examples in the near future? Would be very useful.

tensorflow_serving/example/hot_reload_grpc_example.py
```diff
@@ -0,0 +1,38 @@
...
+
+  request.config.CopyFrom(model_server_config)
+
+  print(request.IsInitialized())
+  print(request.ListFields())
+
+  responese = stub.HandleReloadConfigRequest(request,10)
```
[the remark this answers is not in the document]
karlschriek: Have you maybe found a way to get the server to pass its current model config to the client by a REST or gRPC request? Currently the client has to have access to the model config file in order to trigger a reload without the existing models being unloaded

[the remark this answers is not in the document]
rhlr: @karlschriek Did you find any way to pass the current model config ?

Miopas:
Thank you for this PR. It works for me. Something noticeable in my running environment is:
* ` model_server_config.model_config_list.CopyFrom(config_list) c` The ending `c` is a typo.
* run `pip install tensorflow-serving-api` before run this script.
* The url for gRPC should be `0.0.0.0:8500`.

PR #1066, opened 2018-08-21
Move tools/bazel.rc to .bazelrc

meteorcloudy:
Bazel 0.17.0 will contain a change for which rc files it accepts.
bazelbuild/bazel@ec83598
bazelbuild/bazel#4502

Old bazel used to read %workspace%/tools/bazel.rc. New bazel will not
read that and instead will only read %workspace%/.bazelrc.


chrisolston:
Hi, Thanks for the PR. We'll make this change as needed from our side as we update the bazel version we support.

PR #1067, opened 2018-08-23
use if_tensorrt to integrate NVIDIA TensorRT

qiaohaijun:
if configure NEED_TENSORRT.  tensorflow serving will integrate TensorRT.

### DIFF
--- a/tensorflow_serving/model_servers/BUILD
+++ b/tensorflow_serving/model_servers/BUILD
@@ -278,10 +278,16 @@ cc_library(
     ],
 )
 
+load("@local_config_tensorrt//:build_defs.bzl", "if_tensorrt",)
+
+
 SUPPORTED_TENSORFLOW_OPS = [
     "@org_tensorflow//tensorflow/contrib:contrib_kernels",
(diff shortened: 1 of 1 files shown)


aaroey:
TensorRT is already added, we're working on updating build scripts and once it's done TRT will be available in TF Serving.

Chris19920210:
Hi, Is the build script be updated now ? I wanna build the TF serving with TensorRT support. Thanks. @aaroey

aaroey:
@Chris19920210 Yes it is, see the [dockerfile](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/Dockerfile.gpu#L30).

Chris19920210:
@aaroey Thanks. I've already pulled and built the dockerfile. And I know that it has installed some packages to support tensorrt. But the tf serving doesn't necessarily support tensorrt ops. And I have to build from the source codes, right?

aaroey:
@Chris19920210 @PetroWu TF Serving links with TRT ops/kernels (i.e. TRTEngineOp) [here](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/BUILD#L296).

PR #1072, opened 2018-08-24
update guide doc for custom source

naurril:
The `FileSystemStoragePathSource` now supports configurable `servable version policy` and reports multiple versions rather than the "latest one version" (which could be incorrect even when the ` servable version policy` is  set as `Latest`).


naurril:
I have changed the commit, please check if it is OK. 
Thank you!

PR #1076, opened 2018-09-05
Expose model config files in Docker to serve multiple models

felixduvallet:
Expose the --model_config_file argument in the dockerized tensorflow
model server by using an environment variable. The model config file
specifies a list of model(s) and path(s) to use during serving.

This enables a single model server to serve multiple models.

For example, to serve two different MNIST models (A and B), provide a
config file such as this one:

    model_config_list: {
      config: {
        name: "mnistA",
        base_path: "/models/mnistA",
        model_platform: "tensorflow"
      },
      config: {
        name: "mnistB",
        base_path: "/models/mnistB",
        model_platform: "tensorflow"
      }
    }

The model server container can then be started by passing the variable
    -e MODEL_CONFIG_FILE=/models/models.conf

Note: Generally, this is more useful to serve to *separate* models.

### DIFF
--- a/tensorflow_serving/tools/docker/Dockerfile
+++ b/tensorflow_serving/tools/docker/Dockerfile
@@ -47,4 +47,9 @@ RUN mkdir -p ${MODEL_BASE_PATH}
 
 # The only required piece is the model name in order to differentiate endpoints
 ENV MODEL_NAME=model
-ENTRYPOINT tensorflow_model_server --port=8500 --rest_api_port=8501 --model_name=${MODEL_NAME} --model_base_path=${MODEL_BASE_PATH}/${MODEL_NAME}
+
+# If non-empty, this config file will be used to specify which model(s) to load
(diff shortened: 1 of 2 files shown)


gautamvasudevan:
Thanks for this PR! One of the downsides of this approach is that every single argument you'd like to pass to the model server binary will need to be exposed as an environment variable. This is pretty unfortunate.

I've submitted an alternative approach [here](https://github.com/tensorflow/serving/commit/585e0668ecf6e6c8b61b2bff5c4fdbb32991ec08) that works a bit more generically for all future arguments you might want to tweak.

Thanks for pointing out this deficiency! Let me know if the submitted change works for you (you'll have to use the nightly to test it out initially).

felixduvallet:
Thanks, I like the generic solution better as well.

lozuwa:
Hey, thanks for the docs. Though do you have any idea on how to add this extra parameter to a docker-compose? 
Of course I have built an image based on tensorflow/serving:gpu-latest with my own model.

PR #1092, opened 2018-09-14
Fixes #1091 - Support gzip encoded requests

### LINKED ISSUE #1091: Support accepting gzipped requests
### Describe the problem the feature is intended to solve
Currently tensorflow-serving does not handle GZipped request bodies in the REST endpoint, which can slow down POSTing large volumes of data to tensorflow-serving. It fails with a JSON parse error if this is the case, even if the correct headers are sent. Networks are fast, sure, but why send ~10 MB JSON bodies when you can send ~100kb ones?

### Describe the solution
If `Content-Encoding: gzip` is sent as a header then the body should be decompressed before parsing. 

### Describe alternatives you've considered
A reverse proxy that decompresses responses before passing them to tf-serving, but this is not an ideal solution and adds overhead.

### Additional context
When encoding an image for inference the resulting JSON can be very large, upwards of 10 megabytes. Gzipped this often is reduced to ~400kb.

orf:
I've never written any C++ nor used Bazel before, so please don't laugh if this is terrible!

I've had a go at supporting gzipped request bodies inside the tensorflow-serving http server: #1091 

I've got no idea if the ticket will be accepted, but I thought it would be fun to have a go at fixing it.

I've also got no idea how to test this, I copied an existing test and made it used zlib (compatible with gzip) to send encoded data.

### DIFF
--- a/tensorflow_serving/model_servers/BUILD
+++ b/tensorflow_serving/model_servers/BUILD
@@ -251,6 +251,7 @@ cc_library(
         "@com_google_absl//absl/strings",
         "@com_googlesource_code_re2//:re2",
         "@org_tensorflow//tensorflow/core:lib",
+        "@com_github_mapbox_gziphpp//:gzip"
     ],
 )
 
--- a/tensorflow_serving/model_servers/http_server.cc
+++ b/tensorflow_serving/model_servers/http_server.cc
@@ -16,6 +16,7 @@ limitations under the License.
 #include <cstdint>
 #include <memory>
 
+#include "gzip/decompress.hpp"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_format.h"
 #include "absl/strings/string_view.h"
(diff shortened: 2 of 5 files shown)


orf:
I signed it!

netfs:
Thanks for your effort and the PR!

wenbozhu:
Sorry for the delayed response. I am working on the tensorflow-serving http server and the proper fix will require a bit more work (to be spec compliant etc). We may not want to depend on the gzip wrapper library either. 

Do you mind filing an issue (with a link to this PR) and I try to get the feature supported soon?

Thanks.

orf:
> Sorry for the delayed response. I am working on the tensorflow-serving http server and the proper fix will require a bit more work (to be spec compliant etc). We may not want to depend on the gzip wrapper library either.
> 
> Do you mind filing an issue (with a link to this PR) and I try to get the feature supported soon?
> 
> Thanks.

I've filed a bug here: #1091 

Yeah this will take more work, I guess this was just a POC more than anything else. I saw that Tensorflow itself has some gzip/zlib decoders, but they seem tightly coupled to files and hard to adapt here.

gautamvasudevan:
With the issue filed, we'll track the work there. Thanks again!

PR #1159, opened 2018-10-24
Add support for listening gRPC over UNIX socket

vtorhonen:
One typical deployment model for Tensorflow Serving is to run it as a sidecar container. With this approach the model is often served over HTTP through a loopback interface. For performance reasons it would make sense to offer a possibility to access Serving over UNIX sockets. This would remove TCP overhead and reduce context switching.

This PR adds a new CLI flag `--grpc_socket_path`. If defined, Serving will listen to a UNIX domain socket at this path. It can be a relative or an absolute path. Note that abstract UNIX sockets are not supported with gRPC. There is an issue about this at https://github.com/grpc/grpc/issues/4677.

### DIFF
--- a/tensorflow_serving/model_servers/server.cc
+++ b/tensorflow_serving/model_servers/server.cc
@@ -293,7 +293,7 @@ Status Server::BuildAndStart(const Options& server_options) {
   LOG(INFO) << "Running gRPC ModelServer at " << server_address << " ...";
   if (!server_options.grpc_socket_path.empty()) {
     LOG(INFO) << "Running gRPC ModelServer at UNIX socket "
-              << server_options.grpc_socket_path << "...";
+              << server_options.grpc_socket_path << " ...";
   }
 
   if (server_options.http_port != 0) {
--- a/tensorflow_serving/model_servers/server.cc
+++ b/tensorflow_serving/model_servers/server.cc
@@ -264,9 +264,8 @@ Status Server::BuildAndStart(const Options& server_options) {
       server_address,
       BuildServerCredentialsFromSSLConfigFile(server_options.ssl_config_file));
   // If defined, listen to a UNIX socket for gRPC.
-  const string grpc_socket_path = server_options.grpc_socket_path;
-  const string grpc_socket_uri = "unix:" + grpc_socket_path;
-  if (!grpc_socket_path.empty()) {
+  if (!server_options.grpc_socket_path.empty()) {
+    const string grpc_socket_uri = "unix:" + server_options.grpc_socket_path;
     builder.AddListeningPort(
       grpc_socket_uri,
       BuildServerCredentialsFromSSLConfigFile(server_options.ssl_config_file));
@@ -292,9 +291,9 @@ Status Server::BuildAndStart(const Options& server_options) {
     return errors::InvalidArgument("Failed to BuildAndStart gRPC server");
   }
   LOG(INFO) << "Running gRPC ModelServer at " << server_address << " ...";
-  if (!grpc_socket_path.empty()) {
+  if (!server_options.grpc_socket_path.empty()) {
     LOG(INFO) << "Running gRPC ModelServer at UNIX socket "
-              << grpc_socket_uri << "...";
+              << server_options.grpc_socket_path << "...";
   }
 
   if (server_options.http_port != 0) {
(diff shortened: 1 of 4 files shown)


netfs:
thanks for the change! i will review this sometime this week.

do you have any benchmarks to show case improvements in latency (or throughput)
when going over unix domain sockets vs tcp/ip for ML inference? i'd be curious to see
the results.

vtorhonen:
Sorry, I had too much fun with Tensorflow and it took a while to get back here. :)

Here are some benchmark results from a project that runs Tensorflow Serving on GPUs on Google Kubernetes Engine. Our client app uses [Applifier/go-tensorflow](https://github.com/applifier/go-tensorflow) to interface with Tensorflow Serving over gRPC. We also built a benchmark tool with the same library. We did two separate runs with the benchmark tool, first by calling Tensorflow Serving over a UNIX domain socket and then by calling it over the default TCP socket. There was a small pause between the runs.

This first graph shows the average rate of successful predictions per second.

pr1159-predictions-per-second

The second graph shows the average endpoint latency seen by the gRPC client.

pr1159-grpc-endpoint-latency

There's a huge difference in latency with the default TCP socket, which is actually why our test run eventually failed to finish. Peak median latency increased from 30 ms to 135 ms, where p99 latency increased from 260 ms to 480 ms.

Unfortunately I don't have a pathological and reproducible example about this.

orktes:
@netfs Do you still need more info?

tensorflow_serving/model_servers/server.cc
```diff
@@ -263,6 +263,14 @@ Status Server::BuildAndStart(const Options& server_options) {
   builder.AddListeningPort(
       server_address,
       BuildServerCredentialsFromSSLConfigFile(server_options.ssl_config_file));
+  // If defined, listen to a UNIX socket for gRPC.
+  const string grpc_socket_path = server_options.grpc_socket_path;
+  const string grpc_socket_uri = "unix:" + grpc_socket_path;
```
netfs: can you please move line 268 inside the if block after you have checked for it being empty?

tensorflow_serving/model_servers/server.cc
```diff
@@ -263,6 +263,14 @@ Status Server::BuildAndStart(const Options& server_options) {
...
+  if (!grpc_socket_path.empty()) {
```
netfs: suggestion: May want to directly do if (server_options.grpc_socket_path.empty()) ?

tensorflow_serving/model_servers/server.cc
```diff
@@ -284,6 +292,10 @@ Status Server::BuildAndStart(const Options& server_options) {
     return errors::InvalidArgument("Failed to BuildAndStart gRPC server");
   }
   LOG(INFO) << "Running gRPC ModelServer at " << server_address << " ...";
+  if (!grpc_socket_path.empty()) {
+    LOG(INFO) << "Running gRPC ModelServer at UNIX socket "
```
netfs: can you please add a test (attempt calling one of the api endpoints, say "predict") to:

serving/tensorflow_serving/model_servers/tensorflow_model_server_test.py

the (test) server listens on tcp+http ports, have it also listen on UDS. and then add
a new test method (say testPredictUDS) to test on UDS. no need to add tests for
each of the API end points. this is mostly to check if grpc/uds is functional.

thanks!

netfs:
> @netfs Do you still need more info?

nope. this looks good. thanks for the change, started to take a look/review!

tensorflow_serving/model_servers/server.cc
```diff
@@ -284,6 +291,10 @@ Status Server::BuildAndStart(const Options& server_options) {
     return errors::InvalidArgument("Failed to BuildAndStart gRPC server");
   }
   LOG(INFO) << "Running gRPC ModelServer at " << server_address << " ...";
+  if (!server_options.grpc_socket_path.empty()) {
+    LOG(INFO) << "Running gRPC ModelServer at UNIX socket "
+              << server_options.grpc_socket_path << "...";
```
netfs: space before the ellipsis. s/.../ .../

tensorflow_serving/model_servers/server.cc
vtorhonen: Yes, good call! I went with that since it aligned so well with previous code on that file.

tensorflow_serving/model_servers/server.cc
vtorhonen: Cheers, fixed!

ndeepesh:
Hi @vtorhonen 
Do you have the code for the benchmark that you performed? I am trying a similar thing and latency (p99) is worse for unix domain socket. Just wanted to understand if you ran your configuration with some specific configurations

PR #1176, opened 2018-11-07
[INTEL MKL]: ADD TFServing+MKL dockerfiles

karthikvadla:
This PR adds MKL dockerfiles for TFServing.


tensorflow_serving/tools/docker/Dockerfile.mkl
```diff
@@ -0,0 +1,58 @@
...
+# The only required piece is the model name in order to differentiate endpoints
+ENV MODEL_NAME=model
+
+# Set inter_op or intra_op threads
+ENV TENSORFLOW_SESSION_PARALLELISM=2
+
+# Create a script that runs the model server so we can use environment variables
+# while also passing in arguments from the docker command line
+RUN echo '#!/bin/bash \n\n\
+tensorflow_model_server --port=8500 --rest_api_port=8501 \
```
dzungductran: Maybe you should add ENV GPRC_PORT 8500, then use it here as --port=${GRPC_PORT}?

tensorflow_serving/tools/docker/Dockerfile.mkl
```diff
@@ -0,0 +1,58 @@
...
+
+# Set where models should be stored in the container
+ENV MODEL_BASE_PATH=/models
+RUN mkdir -p ${MODEL_BASE_PATH}
+
+# The only required piece is the model name in order to differentiate endpoints
+ENV MODEL_NAME=model
+
+# Set inter_op or intra_op threads
+ENV TENSORFLOW_SESSION_PARALLELISM=2
```
dzungductran: Why is this set to 2? Isn't this the number of cores?

karthikvadla:
```
CORES_PER_SOCKET=`lscpu | grep "Core(s) per socket" | cut -d':' -f2 | sed "s/ //g"`
SOCKETS=`lscpu | grep "Socket(s)" | cut -d':' -f2 | sed "s/ //g"`
TOTAL_PHYSICAL_CORES=$((${CORES_PER_SOCKET} * ${SOCKETS} ))
HALF_CORES=$(( ${TOTAL_PHYSICAL_CORES} / 2))

ENV TENSORFLOW_SESSION_PARALLELISM=${HALF_CORES}
```
Above snippet will assign half of physical cores by default, if that is okay. I will update.

dzungductran: Can you just do getconf _NPROCESSORS_ONLN ?  This is less error prone. I just don't like the grep and sed.

karthikvadla: Oh i see that is portable, but it will give you total cores=physical+logical cores. We have to divide twice to get the actual physical core counts for the half. 🤔

tensorflow_serving/tools/docker/Dockerfile.devel-mkl
```diff
@@ -0,0 +1,110 @@
...
+    git remote add upstream https://github.com/tensorflow/serving.git && \
+    if [ "${TF_SERVING_VERSION_GIT_COMMIT}" != "head" ]; then git checkout ${TF_SERVING_VERSION_GIT_COMMIT} ; fi
+
+
+FROM base_build as binary_build
+
+# Build, and install TensorFlow Serving
+# Building with -march=sandybridge will allow the container to run on as many platforms as possible;
+# Intel(R) MKL-DNN will still take advantage of all modern CPU instructions at runtime.
+ARG TF_SERVING_BUILD_OPTIONS="--define=using_mkl=true --copt=-march=native --copt=-O3 --host_copt=-march=native -c opt"
```
gautamvasudevan: These args should be just --config=mkl and specified in the bazel.rc file, so it can be built easily outside of the dockerfile as well.

tensorflow_serving/tools/docker/Dockerfile.mkl
dzungductran: Oh, what happens when there is no logical cores?  I don't think it is a good idea to divide twice. 
Also, total cores including logical cores should be fine for number of threads it can handle, so I don't know why you can't just do half of that only.

tensorflow_serving/tools/docker/Dockerfile.mkl
gautamvasudevan: Why not just bind ports in docker to remap them?

gautamvasudevan:
Something else we need is a model (perhaps half plus two) that we can send a rest query to and verify that it is working, and that the mkl optimizations are being exercised (such that it would fail if it were run against a standard server).

karthikvadla:
yeah i tested above docker files with `half plus two` model before creating this PR. What you are expecting exactly here, any README with instructions or any script to run model against these dockerfile binaries?

tensorflow_serving/tools/docker/Dockerfile.mkl
gautamvasudevan: yeah, not sure this needs to be changed, but if there's a good reason it should be reflected in all serving dockerfiles.

gautamvasudevan:
Basically something like a unit test would be handy. For the GPU model, we test it against a half_plus_two model that binds the ops to the GPU so it will fail if you run it on the CPU. Unfortunately there isn't a readme or any specific instructions here as it hasn't come up yet.

tensorflow_serving/tools/docker/Dockerfile.mkl
karthikvadla: Can you explain more about this `bind ports in docker to remap them`? I mean what are you expecting here.

gautamvasudevan: I'm expecting no change, this was a response to @dzungductran's suggestion.

tensorflow_serving/tools/docker/Dockerfile.mkl
gautamvasudevan: I think your snippet seems right given all the socket combinations possible. You should definitely add a comment about how you arrived at this value :)

karthikvadla: yes sure.

tensorflow_serving/tools/docker/Dockerfile.devel-mkl
karthikvadla: or else i will document to pass this is `--build-arg TF_SERVING_BUILD_OPTIONS="--define=using_mkl=true --copt=-march=native --copt=-O3 --host_copt=-march=native -c opt"` during `docker build`. 

If you want to stick to `.bazelrc` route let me know. I don't see any any `.bazelrc` related stuff in your actual `Dockerfile`. I'm very new `bazel` related stuff, if you guide me through it happy to work. 
@gautamvasudevan

gautamvasudevan: Yes, see how it looks in the Dockerfile.devel piece. You should be able to use `--config=mkl --config=nativeopt` instead.

karthikvadla: Sure will work on it. All i have to do is use `ARG TF_SERVING_BUILD_OPTIONS="--config=mkl"` which does bazel build with `MKL` optimizations. 

I don't see anywhere you building `tensorflow` in the dockerfile. Where  is that happening?
 and which version of `tensorflow` is used by `tensorflow-seving`? @gautamvasudevan

gautamvasudevan: Tensorflow is built as a [dependency in bazel](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/BUILD). The version used for releases maps 1:1 usually (so 1.12 TF Serving uses 1.12 of TF). However, for the master build, we build off of the latest green build of TF (which is usually at or very near master).

karthikvadla:
@gautamvasudevan i made few updates to the dockerfile, Can you please point me to GPU unit tests you are talking about? i will take that a reference to write for CPU or any example unit should be good.

tensorflow_serving/tools/docker/Dockerfile.mkl
```diff
@@ -44,28 +44,37 @@ RUN mkdir -p ${MODEL_BASE_PATH}
...
-export KMP_BLOCKTIME=1 \n\
-export KMP_SETTINGS=1 \n\
-export KMP_AFFINITY=granularity=fine,verbose,compact,1,0 \n\
+export OMP_NUM_THREADS=${OMP_NUM_THREADS} \n\
+export KMP_BLOCKTIME=${KMP_BLOCKTIME} \n\
+export KMP_SETTINGS=${KMP_SETTINGS} \n\
+export KMP_AFFINITY=${KMP_AFFINITY} \n\
 tensorflow_model_server --port=8500 --rest_api_port=8501 \
---tensorflow_session_parallelism=${HALF_CORES} \
+--tensorflow_session_parallelism=${TENSORFLOW_SESSION_PARALLELISM} \
```
dzungductran: Where does TENSORFLOW_SESSION_PARALLELISM come from? Is it from environment?  If it is, then you are asking the user to figure this out which seems error prone.

karthikvadla: By default it will assign `2` which is current functionality. We are explicitly explaining user to assign 1/4th of physical cores to get optimal performance (all are mentioned in comments). If we dynamically assign values, that is going to fail on kubernetes with resource constraint issues.

karthikvadla: @dzungductran https://github.com/tensorflow/serving/blob/75af4a0c813a0edc965f4031a324a3df28dc2538/tensorflow_serving/tools/docker/Dockerfile.mkl#L49-L69

dzungductran:
I guess you could, but there already existing dockerfiles for other optimization like GPU in the existing directory. I think you could create a Dockerfile.ppc.

gautamvasudevan:
> @gautamvasudevan i made few updates to the dockerfile, Can you please point me to GPU unit tests you are talking about? i will take that a reference to write for CPU or any example unit should be good.

Sorry it took a bit - I've [exposed a unit test setup](https://github.com/tensorflow/serving/commit/44318969563f951d91b22f7cb5f827873abd6d99) that you can follow for testing this docker image.

karthikvadla:
> > @gautamvasudevan i made few updates to the dockerfile, Can you please point me to GPU unit tests you are talking about? i will take that a reference to write for CPU or any example unit should be good.
> 
> Sorry it took a bit - I've [exposed a unit test setup](https://github.com/tensorflow/serving/commit/44318969563f951d91b22f7cb5f827873abd6d99) that you can follow for testing this docker image.

Thanks @gautamvasudevan . I tried `mkl` tests based on this reference .
My `bazel test` fails with error.
```
bazel test :unittest_Dockerfile.mkl --test_arg=tensorflow/serving:latest-mkl --test_output=streamed --verbose_failures
WARNING: Streamed test output requested. All tests will be run locally, without sharding, one at a time
ERROR: Skipping ':unittest_Dockerfile.mkl': no such package '': BUILD file not found on package path
ERROR: no such package '': BUILD file not found on package path
INFO: Elapsed time: 0.052s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```

Wondering is this typo?
`https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/tests/dockerfile_devel_test.sh#L24`
Because in `BUILD` file it is named 
https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/tests/BUILD#L34 as this. 

i tried this way too
```
bazel test :unittest_dockerfile_mkl --test_arg=tensorflow/serving:latest-mkl --test_output=streamed --verbose_failures
WARNING: Streamed test output requested. All tests will be run locally, without sharding, one at a time
ERROR: Skipping ':unittest_dockerfile_mkl': no such package '': BUILD file not found on package path
ERROR: no such package '': BUILD file not found on package path
INFO: Elapsed time: 0.054s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```
Still same error. Should i add anything to bazel path?
I have my images locally built too
```
$ docker images | grep tensorflow/serving
tensorflow/serving                                     latest-mkl          4e025604bee6        34 minutes ago      519MB
tensorflow/serving                                     latest-devel-mkl    a81e72c68de1        35 minutes ago      5.04GB
```

gautamvasudevan:
You just need to provide the correct path to the BUILD file. The instruction assumes you’re running from within the tests/ folder.

karthikvadla:
> You just need to provide the correct path to the BUILD file. The instruction assumes you’re running from within the tests/ folder.

https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/tests/dockerfile_devel_test.sh#L24 , what about the unit test name . It should be same as you mention in BUILD while running `bazel test`?

karthikvadla:
Yep.  did as below
```
bazel test :unittest_dockerfile_devel_mkl --test_arg=tensorflow/serving:latest-devel-mkl --test_output=streamed --verbose_failures
```
```
Successfully queries model. Response:\n{
    "predictions": [2.5, 3.0, 4.5
    ]
}
Stopping TF ModelServer
a3117789dcce
Clean up TMPDIR:/tmp/tmp.fEmSGBfwFX
Target //tensorflow_serving/tools/docker/tests:unittest_dockerfile_devel_mkl up-to-date:
  bazel-bin/tensorflow_serving/tools/docker/tests/unittest_dockerfile_devel_mkl
INFO: Elapsed time: 6.940s, Critical Path: 6.71s
INFO: 1 process: 1 local.
INFO: Build completed successfully, 2 total actions
//tensorflow_serving/tools/docker/tests:unittest_dockerfile_devel_mkl    PASSED in 6.7s

Executed 1 out of 1 test: 1 test passes.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see whINFO: Build completed successfully, 2 total actions
```. Will push changes

karthikvadla:
Closing this PR. New one opened with all changes #1203

PR #1180, opened 2018-11-09
Fix default path of metrics endpoint

mrkm4ntr:
Fix to follow [this sentence](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/config/monitoring_config.proto#L12).

### DIFF
--- a/tensorflow_serving/model_servers/http_server.cc
+++ b/tensorflow_serving/model_servers/http_server.cc
@@ -83,16 +83,16 @@ net_http::HTTPStatusCode ToHTTPStatusCode(const Status& status) {
 }
 
 void ProcessPrometheusRequest(PrometheusExporter* exporter,
-                              const PrometheusConfig& prometheus_config,
+                              const string& path,
                               net_http::ServerRequestInterface* req) {
   std::vector<std::pair<string, string>> headers;
   headers.push_back({"Content-Type", "text/plain"});
   string output;
   Status status;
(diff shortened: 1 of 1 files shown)


tensorflow_serving/model_servers/http_server.cc
```diff
@@ -83,20 +83,11 @@ net_http::HTTPStatusCode ToHTTPStatusCode(const Status& status) {
...
 
 void ProcessPrometheusRequest(PrometheusExporter* exporter,
-                              const PrometheusConfig& prometheus_config,
                               net_http::ServerRequestInterface* req) {
   std::vector<std::pair<string, string>> headers;
   headers.push_back({"Content-Type", "text/plain"});
   string output;
-  Status status;
-  // Check if url matches the path.
-  if (req->uri_path() != prometheus_config.path()) {
```
mrkm4ntr: I think this check is redundant.

netfs: (firstly sorry for the delayed response!)

yes this check is redundant, but having it here -- allows us to check for
potential misrouting either due to a bug in HTTP server handler code,
or handler registration code (in CreateAndStartHttpSever() below). i'd
like to keep it for now. wdyt?

mrkm4ntr: I see. I changed to keep it.

PR #1185, opened 2018-11-13
Server support for modelconf monitoring and model(s) load/unload

vitalyli:
This will make model server to monitor for any changes inside modelconf proto file; if new model type is added, it will load those that were added; similarly if something is removed, it will unload those.
The new server parameter is: modelconf_poll_wait_seconds, which defaults to 0. This function will Not be activated unless a value greater than 0 is supplied during server startup. If value greater than 0 is given, the monitor will start and perform its function at specified frequency in seconds.


vitalyli:
@chrisolston Hi Chris, could you review this patch? we thought that everything was fine as it reported successful load/re-load of the model when we were modifying the modelconf proto. However, later we ran into segmentation fault on the first attempt to do inference.

PR #1187, opened 2018-11-14
Modelconf auto-manager 

vitalyli:
This will make model server to monitor for any changes inside modelconf proto file; if new model type is added, it will load those that were added; similarly if something is removed, it will unload those.
The new server parameter is: modelconf_poll_wait_seconds, which defaults to 0. This function will Not be activated unless a value greater than 0 is supplied during server startup. If value greater than 0 is given, the monitor will start and perform its function at specified frequency in seconds.

### DIFF
--- a/tensorflow_serving/model_servers/main.cc
+++ b/tensorflow_serving/model_servers/main.cc
@@ -102,6 +102,10 @@ int main(int argc, char** argv) {
                        &options.file_system_poll_wait_seconds,
                        "interval in seconds between each poll of the file "
                        "system for new model version"),
+      tensorflow::Flag("modelconf_poll_wait_seconds",
+                       &options.modelconf_poll_wait_seconds,
+                       "activates model_config_file monitor in seconds; "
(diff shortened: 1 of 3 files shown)


gautamvasudevan:
Is there a reason to poll vs. just sending a request to reload the config to the ModelServer (the latter exists already)? Polling systems tend to lead to having to avoid race conditions based on configuration options, while sending an imperative command when you want the server to reload the config (a possibly expensive operation) gives you a lot more flexibility (including implementing a simple polling system externally if that's ideal for you).

vitalyli:
Thanks for your review. What I showed here is working Ok for now. I may be able to get to it again and re-do it with REST API. Architecturally it's probably cleaner that way anyway.

gautamvasudevan:
Great! I think for this project, building upon the REST interface is the right way to go. I'll close this PR - if you get the chance to add support on the REST endpoint, feel free to open a new PR.

Thanks!

PR #1225, opened 2018-12-27
Making resnet_client.py python3 compliant:

rbahumi:
I did not manage to use  the resnet_client.py example with python 3.
I suggest the following modification so the file is also python3 compliant:

1. Using a bytes-like object in the base64.b64encode() method
2. Modified xrange=>range. 'xrange' keyword is not python3 complient and a generator is not required for such short sequence.

### DIFF
--- a/tensorflow_serving/example/resnet_client.py
+++ b/tensorflow_serving/example/resnet_client.py
@@ -48,18 +48,18 @@ def main():
   dl_request.raise_for_status()
 
   # Compose a JSON Predict request (send JPEG image in base64).
-  predict_request = '{"instances" : [{"b64": "%s"}]}' % base64.b64encode(
+  predict_request = b'{"instances" : [{"b64": "%s"}]}' % base64.b64encode(
       dl_request.content)
 
   # Send few requests to warm-up the model.
-  for _ in xrange(3):
+  for _ in range(3):
     response = requests.post(SERVER_URL, data=predict_request)
     response.raise_for_status()
 
   # Send few actual requests and report average latency.
   total_time = 0
   num_requests = 10
-  for _ in xrange(num_requests):
+  for _ in range(num_requests):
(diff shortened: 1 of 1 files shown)


rbahumi:
I signed it!

tensorflow_serving/example/resnet_client.py
```diff
@@ -48,18 +48,18 @@ def main():
   dl_request.raise_for_status()
 
   # Compose a JSON Predict request (send JPEG image in base64).
-  predict_request = '{"instances" : [{"b64": "%s"}]}' % base64.b64encode(
+  predict_request = b'{"instances" : [{"b64": "%s"}]}' % base64.b64encode(
```
gautamvasudevan: Why is this necessary?

gautamvasudevan: Also this seems to treat the entire string as bytes rather than just the base64 encoding.

rbahumi: It is necessary for python3, it does not work otherwise.

rbahumi: In python3, the default is unicode which does not comply with base64.
[Here is a link to a related stack overflow issue.](https://stackoverflow.com/questions/8908287/why-do-i-need-b-to-encode-a-python-string-with-base64)

gautamvasudevan: Shouldn't just the b64 data be encoded?

```python
jpeg_bytes = dl_request.content.encode("utf-8")
predict_request = '{"instances" : [{"b64": "%s"}]}' % base64.b64encode(jpeg_bytes)
```

gautamvasudevan:
Closing - I've made the necessary changes.

rbahumi:
Sorry for not responding. I wanted to verify/run your suggestion before commenting and didn't got to it. Thank you.

Issue #1237, opened 2019-01-15
Split tensorflow_session_parallelism for optimal performance 


karthikvadla:
## Feature Request
Splitting `tensorflow_session_parallelism` into `intra_op_parallelism_threads` and `inter_op_parallelism_threads`.

**TENSORFLOW_SESSION_PARALLELISM** is currently responsible for setting both inter-op and intra-op-parallelism-threads in TensorFlow Serving. This is limiting since these two TF parameters usually have different optimal values.

Please read out this [section](
https://www.tensorflow.org/guide/performance/overview#tuning_mkl_for_the_best_performance)

*intra_op_parallelism_threads:* Setting this equal to the number of physical cores is recommended. Setting the value to 0, which is the default, results in the value being set to the number of logical cores - this is an alternate option to try for some architectures. This value and OMP_NUM_THREADS(while using MKL-DNN) should be equal.

*inter_op_parallelism_threads:* Setting this equal to the number of sockets is recommended. Setting the value to 0, which is the default, results in the value being set to the number of logical cores.

Relevant [code](https://github.com/tensorflow/serving/search?q=tensorflow_session_parallelism&unscoped_q=tensorflow_session_parallelism)

### Describe the solution
Split this flag into new flags called  `intra_op_parallelism_threads` and `inter_op_parallelism_threads` https://github.com/tensorflow/serving/blob/5c7740fc3d8d5c017643a8cc40a7202717b10dd6/tensorflow_serving/model_servers/main.cc#L114

And assigning respective values here
https://github.com/tensorflow/serving/blob/ba8e0a0598ebaf8f73b31652c20c76182e50dbcc/tensorflow_serving/model_servers/server.cc#L225-L229

And update any other relevant code base.

### Alternative
Right now we are assigning 1/4th of physical cores available as a value [here](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/Dockerfile.mkl#L84) with `docker run`

karthikvadla:
@gautamvasudevan 👆 . Can you please assign it to right contact.

gautamvasudevan:
Good idea!

PR #1253, opened 2019-02-05
[FEATURE]: Split tensorflow_session_parallelism for optimal performance

karthikvadla:
Related to #1237 

This PR will add new flags `tensorflow_inter_op_parallelism` and `tensorflow_intra_op_parallelism` to configure values separately. 

To avoid any issues still we support backward compatibility for `tensorflow_session_parallelism`.

Logic goes as below

if either `tensorflow_inter_op_parallelism` OR `tensorflow_intra_op_parallelism` are set, then use those values and ignore `tensorflow_session_parallelism`; if `tensorflow_session_parallelism` is set as well, generate a error message. If neither `tensorflow_inter_op_parallelism` or `tensorflow_intra_op_parallelism` is set, then use `tensorflow_session_parallelism`.

**NOTE:** we'd like the single `tensorflow_session_parallelism` variable to be deprecated in favor of the two `tensorflow_inter_op_parallelism` and `tensorflow_intra_op_parallelism` variables that TensorFlow supports. 

Please let us know your thoughts.

### DIFF
--- a/tensorflow_serving/tools/docker/Dockerfile.mkl
+++ b/tensorflow_serving/tools/docker/Dockerfile.mkl
@@ -81,21 +81,18 @@ ENV KMP_SETTINGS=1
 ENV KMP_AFFINITY='granularity=fine,verbose,compact,1,0'
 ENV MKLDNN_VERBOSE=0
 
-# You can set either TENSORFLOW_SESSION_PARALLELISM or
-# TENSORFLOW_INTRA_OP_PARALLELISM and TENSORFLOW_INTER_OP_PARALLELISM envs separately.
-# Please comment or uncomment as required.
-# If you want to set TENSORFLOW_SESSION_PARALLELISM, uncomment respective env and line in bash script below
-# and comment out TENSORFLOW_INTRA_OP_PARALLELISM and TENSORFLOW_INTER_OP_PARALLELISM along with respective
-# lines in below bash script.
+# Recommended settings, may vary depending on the model.
+# Set TENSORFLOW_INTRA_OP_PARALLELISM=#No.of Physical cores
+# Set TENSORFLOW_INTER_OP_PARALLELISM=#No.of Sockets
+# For more information vist below reference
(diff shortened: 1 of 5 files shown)


karthikvadla:
@gautamvasudevan Please take a look when you get time, let me know your thoughts.

tensorflow_serving/tools/docker/Dockerfile.mkl
```diff
@@ -81,11 +81,23 @@ ENV KMP_SETTINGS=1
 ENV KMP_AFFINITY='granularity=fine,verbose,compact,1,0'
 ENV MKLDNN_VERBOSE=0
 
-ENV TENSORFLOW_SESSION_PARALLELISM=2
+# You can set either TENSORFLOW_SESSION_PARALLELISM or
```
gautamvasudevan: Is there any reason why someone would want to keep session parallelism? Given they have to modify the dockerfile regardless (and that there are a ton of flags not pre-configured for a user), it's not clear we need to bother with this.

gautamvasudevan:
Thanks for putting this together!

tensorflow_serving/tools/docker/Dockerfile.mkl
karthikvadla: We don't need this be configured any more. Just for user to know i added those comments. But can remove them.

karthikvadla: You want me to get rid off that ?

gautamvasudevan: Yeah, I'd remove them in that case. Makes it cleaner/easier to understand.

karthikvadla: Sure will update the changes.

karthikvadla:
Hi @vinuraja, did you get a chance to look at PR. Please let me know for any changes. :)

karthikvadla:
Hi @gautamvasudevan/ @vinuraja, what are the chances that this will be pulled into next release (1.13) ?

gautamvasudevan:
Working on that now

PR #1281, opened 2019-03-15
Update bazel_in_docker.sh

guotong1988:


netfs:
this is fixed in #1138 and available in TF 1.12 (and upward) versions.

your PR seems to be against older TF 1.9 release, and this being a non-security/critical fix, will not get cherry-picked in the 1.9 branch. can you please use TF 1.12 or later releases?

also this script was renamed to `tools/run_in_docker.sh`

guotong1988:
I meet the problem [here](https://github.com/tensorflow/tensorflow/issues/24828) and downgrade to 1.9

netfs:
> I meet the problem [here](https://github.com/tensorflow/tensorflow/issues/24828) and downgrade to 1.9

hmm. i think your best bet is to get your stuff working on TF Serving version >= 1.12 -- as this change already exists over there. and as i mentioned, our policies do not allow cherry picking this change onto 1.9 release (as it's not in the critical/security-fix bucket). sorry about that.

PR #1302, opened 2019-04-08
fix build error

louiswi:
fix build error on cctz

### DIFF
--- a/tensorflow_serving/workspace.bzl
+++ b/tensorflow_serving/workspace.bzl
@@ -40,8 +40,8 @@ def tf_serving_workspace():
   # ===== CCTZ (Time-zone framework). Used by ABSL.
   native.http_archive(
       name = "com_googlesource_code_cctz",
-      strip_prefix = "cctz-master",
-      url = "https://github.com/google/cctz/archive/master.zip",
+      strip_prefix = "cctz-2.2",
+      url = "https://github.com/google/cctz/archive/v2.2.zip",
   )
 
   # ===== RapidJSON (rapidjson.org) dependencies =====


louiswi:
I signed it!

louiswi:
I was not building on the master, I was building it on the r1.8 branch. When I built failed, I checked the ctzz project and found that there was a new version released. I guess the new version was a problem, so I switched to the old 2.2 version and the compiler can pass.
error log: /root/.cache/bazel/_bazel_root/e53bbb0b0da4e26d24b415310219b953/external/com_googlesource_code_cctz/BUILD:24:1: no such target '@bazel_tools//platforms:ios': target 'ios' not declared in package 'platforms' (did you mean 'os'?) defined by /root/.cache/bazel/_bazel_root/e53bbb0b0da4e26d24b415310219b953/external/bazel_tools/platforms/BUILD and referenced by '@com_googlesource_code_cctz//:ios'

netfs:
> I was not building on the master but building it on the r1.8 branch. When I built failed, I checked the cctz project and found that there was a new version released. I guess the new version was a problem, so I switched to the old 2.2 version and the compiler can pass.
> error log: /root/.cache/bazel/_bazel_root/e53bbb0b0da4e26d24b415310219b953/external/com_googlesource_code_cctz/BUILD:24:1: no such target '@bazel_tools//platforms:ios': target 'ios' not declared in package 'platforms' (did you mean 'os'?) defined by /root/.cache/bazel/_bazel_root/e53bbb0b0da4e26d24b415310219b953/external/bazel_tools/platforms/BUILD and referenced by '@com_googlesource_code_cctz//:ios'

so this change needs to be on the 1.8 release branch. unfortunately, this is not a critical
fix nor a security fix -- and hence i cannot cherrypick this in the 1.8 release. any reason
why you cannot use the latest (1.13) release? or a release before that (1.12) -- both 
should work IMO.

thanks!

PR #1339, opened 2019-05-06
Allow serving the only signature in the graph even if it is not the default serving signature. This will allow users to not explicitly specify the signature name when they have only one signature.

tensorflow-copybara:
Allow serving the only signature in the graph even if it is not the default serving signature. This will allow users to not explicitly specify the signature name when they have only one signature.


netfs:
obsolete change.

PR #1435, opened 2019-09-04
Add shift to consume script name before passing to application

dmc5179:
The Dockerfile entrypoint /usr/bin/tf_serving_entrypoint.sh calls tensorflow_model_server with "$@" which results in an error message from tensorflow_model_server that "/usr/bin/tf_serving_entrypoint.sh" is not a supported option. This is because "$@" passes the first parameter which is the name of the script itself. To resolve this simply add the shift command to consume the first argument and then pass "$@" to tensorflow_model_server.

### DIFF
--- a/tensorflow_serving/tools/docker/Dockerfile
+++ b/tensorflow_serving/tools/docker/Dockerfile
@@ -51,6 +51,7 @@ ENV MODEL_NAME=model
 # Create a script that runs the model server so we can use environment variables
 # while also passing in arguments from the docker command line
 RUN echo '#!/bin/bash \n\n\
+shift \n\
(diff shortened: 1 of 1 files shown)


dmc5179:
@googlebot I signed it!

netfs:
the fix looks correct, but i am puzzled why the instructions here:

https://github.com/tensorflow/serving#serve-a-tensorflow-model-in-60-seconds

do not yield any error. do you know why?

dmc5179:
This was awhile ago but IIRC I was troubleshooting other issues with my model and noticed this error in the logs. I don't think the error actually causes the process/container to fail. It simply complains that the first option is no good and moves onto the next one. But again, this was awhile ago so I'd have to run through it again and see what happens.

Issue #1468, opened 2019-10-21
Getting error following steps in 'Building Standard TensorFlow ModelServer'


peter-mourfield:
## URL(s) with the issue:

https://www.tensorflow.org/tfx/serving/serving_advanced

## Description of issue (what needs changing):

When I attempt to execute  the 
```
tools/run_in_docker.sh python tensorflow_serving/example/mnist_saved_model.py \
  --training_iteration=100 --model_version=1 /tmp/mnist
```
step in the in the tutorial I get the this error:

```
serving % tools/run_in_docker.sh python tensorflow_serving/example/mnist_saved_model.py \
  --training_iteration=100 --model_version=1 /tmp/mnist
== Pulling docker image: tensorflow/serving:nightly-devel
nightly-devel: Pulling from tensorflow/serving
22e816666fd6: Already exists 
079b6d2a1e53: Already exists 
11048ebae908: Already exists 
c58094023a2e: Already exists 
e9d1145448f7: Pull complete 
3b2b266356de: Pull complete 
9f9b2b982b72: Pull complete 
ede8854b3a01: Pull complete 
7bb55a638df9: Pull complete 
bdd9b510b8a7: Pull complete 
90a5454f6928: Pull complete 
1941316fdbd3: Pull complete 
c9c9a434ee49: Pull complete 
Digest: sha256:3b52152115c73a6be79a86cda94c4c94569df9b490a3e40c2530d5a9a007afac
Status: Downloaded newer image for tensorflow/serving:nightly-devel
docker.io/tensorflow/serving:nightly-devel
== Running cmd: sh -c 'cd /Users/***REMOVED***/GitHub/serving; python tensorflow_serving/example/mnist_saved_model.py --training_iteration=100 --model_version=1 /tmp/mnist'
Traceback (most recent call last):
  File "tensorflow_serving/example/mnist_saved_model.py", line 39, in <module>
    tf.app.flags.DEFINE_integer('training_iteration', 1000,
AttributeError: 'module' object has no attribute 'app'
```
I am running this on Mac OS Catalina w/ 8GB RAM allocated to the Docker Engine.

Nochange1001:
same error

suenpun:
same error +1

Nochange1001:
Would you please tell me the version of tensorflow ?

suenpun:
tensorflow             1.8.0
tensorflow-serving-api 1.12.0

in python2.7

[...]

PR #1470, opened 2019-10-23
Update serving_advanced.md

peter-mourfield:
From this issue: https://github.com/tensorflow/serving/issues/1468

By default run_in_docker.sh pulls the tensorflow/serving:nightly-devel image. This image uses TensorFlow 2.0. The mnist_saved_model.py does not use TensorFlow 2.0. I updated the documentation to have run_in_docker.sh explicitily specify which image to pull (1.14.0-devel) in this case. After making this change the examples run as described in the rest of the document.

### DIFF
--- a/tensorflow_serving/g3doc/serving_advanced.md
+++ b/tensorflow_serving/g3doc/serving_advanced.md
@@ -51,14 +51,14 @@ rm -rf /tmp/models
 Train (with 100 iterations) and export the first version of model:
 
 ```shell
-tools/run_in_docker.sh python tensorflow_serving/example/mnist_saved_model.py \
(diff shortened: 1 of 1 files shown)


peter-mourfield:
I did want to say that if there is a different, preferred way of resolving the issue of the TFX Learning Examples not working (like updating the mnist_saved_model.py to support TF 2.0) I'm happy to do it. I just wanted to enter this issue to get the conversation started :)

michaelschiff:
@peter-mourfield I am running into the same problems - what do you think about using  `compat.v1` as an in-between for re-writting the examples for TensorFlow2.  I've gotten the mnist example working in this PR https://github.com/tensorflow/serving/pull/1486

peter-mourfield:
@michaelschiff looks like a better fix to me. I'll close this PR

Issue #1475, opened 2019-11-01
Python Examples Don't Work with Tensorflow 2.0


benhoff:
### Describe the problem
The python examples break in Tensorflow 2. They rely on `tf.contrib.util.make_tensor_proto`. `tf.contrib` is gone. Looks like this is top level now, see here: https://www.tensorflow.org/api_docs/python/tf/make_tensor_proto

I also think that the examples rely on easy access to a pretrained resnet. See [here](https://github.com/tensorflow/serving/blob/793fd90ee41ac34fa4c9261eef2d2c908dca9735/tensorflow_serving/example/resnet_client.py#L23) for example. However, if you follow the link, https://github.com/tensorflow/models/tree/master/official/resnet#pre-trained-model, and then click around, it's not clear how to download a pretrained resnet at all.

gowthamkpr:
@benhoff I am closing this issue as it is not related to Tensorflow Serving. 

Please post this issue in [tensorflow/tensorflow](https://github.com/tensorflow/tensorflow) as it is most aptly suited to this issue. Thanks!

benhoff:
@gowthamkpr I'm talking about the Tensor flow Serving python examples.

https://github.com/tensorflow/serving/blob/master/tensorflow_serving/example/mnist_client.py

vakker:
@benhoff This is an issue in Tensorflow Serving and not related to tensorflow/tensorflow.

E.g. following the example from [here](https://www.tensorflow.org/tfx/serving/serving_basic) throws the error for the MNIST training after running `tools/run_in_docker.sh python tensorflow_serving/example/mnist_saved_model.py  /tmp/mnist `:
```
== Pulling docker image: tensorflow/serving:nightly-devel
nightly-devel: Pulling from tensorflow/serving
7ddbc47eeb70: Pull complete 
...
25593ff628c4: Pull complete 
Digest: sha256:a13ee93d1ff9ff67dfec586965016f10d30008f4d80688ab3c01650de9d119fb
Status: Downloaded newer image for tensorflow/serving:nightly-devel
docker.io/tensorflow/serving:nightly-devel
== Running cmd: sh -c 'cd /home/user/workspace/tools/serving; python tensorflow_serving/example/mnist_saved_model.py /tmp/mnist'
Traceback (most recent call last):
  File "tensorflow_serving/example/mnist_saved_model.py", line 36, in <module>
    from tensorflow.contrib import lookup as contrib_lookup
ImportError: No module named contrib
```

[...]

PR #1477, opened 2019-11-02
remove obsolete method call from python example

benhoff:
The other Tensorflow Serving Python examples are still broken.

This and other issues are addressed in the below issue.
Ref: https://github.com/tensorflow/serving/issues/1475


netfs:
fixed in b052238fe

PR #1481, opened 2019-11-06
fixed small error in predict request example

jkthompson:
Found a small error in this tutorial example.  Looks like 3870ba59a764d859fc137a8363588c94906e0f5f fixed a similar error recently, but it occurred in two places.


jkthompson:
File was moved to tfx repo and issue resolved via different PR.

PR #1506, opened 2019-12-07
Don't unload versions in config under specific-version policy when the version directory missed.

jianpingzju:
Under specific-version policy, when the version directory is deleted by mistake or something wrong with the filesystem, the desired version in config would be unloaded, which is a disaster for online service.
So the feature of this pull request is an advice that don't unload versions in config under specific-version policy even when the version directory missed.


jianpingzju:
without cla

PR #1516, opened 2019-12-13
update error codes proto reference

figroc:
`error_codes.proto` in tensorflow project has been moved from `tensorflow/core/lib/core` into `tensorflow/core/protobuf`.


figroc:
CI checks seem jammed.

netfs:
fixed in ab475bf6c

PR #1518, opened 2019-12-18
Refresh aspired servables/versions following config update

njhill:
Currently when the configured model list is updated via a call to `handleReloadConfigRequest`, the request thread blocks until any newly added models become available.

Their availability however depends on the filesystem polling thread rescanning the filesystem at some periodic interval, meaning that there's an arbitrary delay before the requested changes actually take effect and the RPC returns.

This problem may not be very noticeable with the default polling interval of 1 second, but seems undesirable for longer intervals and in particular makes API-based dynamic reconfiguration incompatible with the `--file_system_poll_wait_seconds=0` setting (in this case all `handleReloadConfigRequest` calls time-out and do not take effect).

### DIFF
--- a/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
+++ b/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
@@ -290,11 +290,9 @@ Status PollFileSystemForConfig(
 
 // Determines if, for any servables in 'config', the file system doesn't
 // currently contain at least one version under its base path.
-Status FailIfZeroVersions(const FileSystemStoragePathSourceConfig& config) {
-  std::map<string, std::vector<ServableData<StoragePath>>>
-      versions_by_servable_name;
-  TF_RETURN_IF_ERROR(
-      PollFileSystemForConfig(config, &versions_by_servable_name));
+Status FailIfZeroVersions(const FileSystemStoragePathSourceConfig& config,
+  std::map<string, std::vector<ServableData<StoragePath>>>&
+      versions_by_servable_name) {
   for (const auto& entry : versions_by_servable_name) {
     const string& servable = entry.first;
     const std::vector<ServableData<StoragePath>>& versions = entry.second;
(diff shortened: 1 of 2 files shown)


njhill:
I have opened this against 1.15 since that's the version we are using, but can rebase on a different branch if needed.

Also apologies in advance for the code, I am not very familiar with C++.

tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
```diff
@@ -398,19 +414,25 @@ Status FileSystemStoragePathSource::PollFileSystemAndInvokeCallback() {
...
+void FileSystemStoragePathSource::LogVersionsAndInvokeCallback(
+    const string& servable,
+    const std::vector<ServableData<StoragePath>>& versions) {
+  for (const ServableData<StoragePath> &version : versions) {
+    if (version.status().ok()) {
+      VLOG(1) << "File-system polling update: Servable:" << version.id()
+              << "; Servable path: " << version.DataOrDie()
+              << "; Polling frequency: "
+              << config_.file_system_poll_wait_seconds();
+    }
```
christisg: Could we move the logging logic into a separate method?
It would make the code would more readable, and allow the flexibility to call CallAspiredVersionsCallback independently.

njhill: @christisg certainly, though that seems kind of unrelated to the fix (this specific logging logic was embedded in a larger method before, and all the other logging I can see is also inline).

njhill:
Thanks @christisg, I've pushed a commit to address your logging comment. I will aim to add unit test coverage when I get a chance... it will take me a bit longer due to unfamiliarity with C++ and the codebase/test framework.

astleychen:
@njhill thanks for reporting this bug. It's painful and took me more than 3 hrs to figure out **REAL** behavior when setting `--file_system_poll_wait_seconds=0` to mitigate GCS bucket class A/B operation request calls in polling. Hope we can see your fixes soon. :)

netfs:
@njhill do you want wrap this PR by adding unit-test as requested by the reviewer?

thanks!

njhill:
@netfs apologies for letting this lag. I am not sure when I will realistically have a chance to do this since I'm especially busy right now and not very familiar with C++ or the testing setup so it would take me a decent chunk of time to do.

Any help with that part would be appreciated!

netfs:
no worries @njhill.

@astleychen do you want to help here and add tests?

PR #1526, opened 2020-01-06
There is one omission for tf.compat.v1

BraveLii:


netfs:
fixed in 06fbf878a98c8bd42

PR #1540, opened 2020-01-27
Export GRPC and REST API port

gogasca:
We are adding support for Google Cloud Run. (Google Cloud Run + Anthos).
The contract for Google Cloud Anthos consists of the ENV Variable PORT. Which we are defining here as PORT. This is used to enable the service and provide a REST endpoint (HTTP). While we can generate a new container based on this image, I would like to expose this option in source container:

```
docker run -p 8502:8502 --mount type=bind,source=/Users/gogasca/Desktop/cloud_run/resnet_v2_fp32_savedmodel_NHWC_jpg/,target=/models/resnet -e PORT=8502 -e MODEL_NAME=resnet --rm -i -t tensorflow/serving
...
2020-01-27 19:20:09.157365: I tensorflow_serving/model_servers/server.cc:378] Exporting HTTP/REST API at:localhost:8502 ...
```

### DIFF
--- a/tensorflow_serving/tools/docker/Dockerfile
+++ b/tensorflow_serving/tools/docker/Dockerfile
@@ -34,12 +34,15 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # Install TF Serving pkg
 COPY --from=build_image /usr/local/bin/tensorflow_model_server /usr/bin/tensorflow_model_server
 
+ENV GRPC_PORT=8500
+ENV PORT=8501
+
 # Expose ports
 # gRPC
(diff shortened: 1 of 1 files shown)


furdarius:
You can always start with custom arguments for ports `docker run tensorflow/serving --port=8550 --rest_api_port=8511`. So, it isn't required to change Dockerfile.

gogasca:
> You can always start with custom arguments for ports `docker run tensorflow/serving --port=8550 --rest_api_port=8511`. So, it isn't required to change Dockerfile.

Thanks for the clarification.

PR #1561, opened 2020-02-25
Add command-line flags to set model memory limit

samikama:
This PR adds a command-line flag to expose total model memory limit parameter to user. For ease of use limit is expressed in MB units and converted to bytes internally when passed to model server.

### DIFF
--- a/tensorflow_serving/model_servers/server.cc
+++ b/tensorflow_serving/model_servers/server.cc
@@ -19,8 +19,10 @@ limitations under the License.
 
 #include <iostream>
 #include <memory>
+#include <limits>
 #include <utility>
 #include <vector>
+#include <algorithm>
 
 #include "google/protobuf/wrappers.pb.h"
 #include "grpc/grpc.h"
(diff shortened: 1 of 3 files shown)


samikama:
@googlebot I signed it!

samikama:
@ebrevdo since you made the most recent commit, any idea who to ping for this PR or opinions on this? Thanks.

tensorflow_serving/model_servers/server.h
```diff
@@ -59,6 +60,8 @@ class Server {
     tensorflow::string batching_parameters_file;
     tensorflow::string model_name;
     tensorflow::int32 max_num_load_retries = 5;
+    tensorflow::int64 total_model_memory_limit_megabytes =
```
netfs, line 63: set default to `0`

tensorflow_serving/model_servers/server.cc
```diff
@@ -285,7 +287,9 @@ Status Server::BuildAndStart(const Options& server_options) {
   options.flush_filesystem_caches = server_options.flush_filesystem_caches;
   options.allow_version_labels_for_unavailable_models =
       server_options.allow_version_labels_for_unavailable_models;
-
+  options.total_model_memory_limit_bytes = std::min(
```
netfs, line 290: update `total_model_memory_limit_bytes` only when `server_options.total_model_memory_limit_megabytes` is not zero (implying someone changed the value via command line). this ensures that default limit continues to be applied as before. your current change assumes that the default is unit64_max.

Issue #1610, opened 2020-04-26
Allow batch size of zero in JSON outputs


peakji:
### Describe the problem the feature is intended to solve
The current version of tensorflow serving (`2.2.0-rc`) does not allow the final output in the row format to have a batch size of zero:
```json
{ "error": "Tensor name: my_layer_1 has invalid batch size: 0" }
```
While in end-to-end systems it is legit to return an empty response.

For example when feeding an image with no cats to a cat-face-detector that returns the coordinates, or passing an article with nothing useful to an information extraction system.

The issue could be further generalized: We should allow the outermost dimension of an output to be zero, since it might not be the *batch* dimension.

By returning `.values` of a **RaggedTensor** along with the `.value_rowids()`, we can put more business logic inside the graph and serve the end users directly with tensorflow serving.

I believe this will open up lots of new possibilities, e.g. serving results of a recall process, filtering input lists, data augmentation, ... 

### Describe the solution
Simply return an empty prediction list instead of an error:
```json
{ "predictions": [] }
```

### Describe alternatives you've considered
We could pad the final outputs to have a batch size of at least 1, but this will result in an object with empty/default values, which may confuse the end users of the serving API: In order to distinguish between a *real* empty result and a list with a single valid item, one must check the first item in the list on every request.

PR #1615, opened 2020-04-28
Allow batch size of zero in row format JSON

peakji:
Fixes https://github.com/tensorflow/serving/issues/1610

### DIFF
--- a/tensorflow_serving/util/json_tensor_test.cc
+++ b/tensorflow_serving/util/json_tensor_test.cc
@@ -1062,6 +1062,35 @@ TEST(JsontensorTest, FromJsonMultipleZeroBatchTensors) {
   TF_EXPECT_OK(CompareJson(json, R"({ "predictions": [] })"));
 }
 
+TEST(JsontensorTest, FromJsonMultipleZeroBatchTensorsErrors) {
+  TensorMap tensormap;
+  ASSERT_TRUE(TextFormat::ParseFromString(R"(
+    dtype: DT_INT32
+    tensor_shape {
+      dim { size: 0 }
+      dim { size: 2 }
+    }
+    )",
+                                          &tensormap["int_tensor"]));
+
+  ASSERT_TRUE(TextFormat::ParseFromString(R"(
+    dtype: DT_STRING
+    tensor_shape {
+      dim { size: 1 }
+      dim { size: 2 }
+    }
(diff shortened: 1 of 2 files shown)


tensorflow_serving/util/json_tensor_test.cc
```diff
@@ -1017,6 +1017,51 @@ TEST(JsontensorTest, FromJsonMultipleNamedTensorsErrors) {
...
+      dim { size: 2 }
+    }
+    )",
+                                          &tensormap["str_tensor"]));
+
+  ASSERT_TRUE(TextFormat::ParseFromString(R"(
+    dtype: DT_FLOAT
+    tensor_shape {
+      dim { }
+      dim { size: 1 }
```
lilao, line 1053: Nit: can you change this size 1 to a different number, to verify that we only require the first dimension to match?

tensorflow_serving/util/json_tensor_test.cc
```diff
@@ -1017,6 +1017,51 @@ TEST(JsontensorTest, FromJsonMultipleNamedTensorsErrors) {
...
+    }
+    )",
+                                          &tensormap["float_tensor"]));
+
+  string json;
+  TF_EXPECT_OK(
+      MakeJsonFromTensors(tensormap, JsonPredictRequestFormat::kRow, &json));
+  TF_EXPECT_OK(CompareJson(json, R"({ "predictions": [] })"));
+}
+
```
lilao, line 1064: Can you also add a test when the first dimension of the first tensor is 0, and that of the second tensor is greater than 0?

tensorflow_serving/util/json_tensor_test.cc
peakji: Updated in https://github.com/tensorflow/serving/pull/1615/commits/1f7aebd906a70ba0fa04105ceee6227960b764f7

tensorflow_serving/util/json_tensor_test.cc
peakji: New test case added in https://github.com/tensorflow/serving/pull/1615/commits/91afd42dab8ce50f86bbf65065dce0c28163314b

peakji:
@lilao Thanks for reviewing my PR, hope we can make it into the `v2.2.0` release, cheers!

Issue #1613, opened 2020-04-27
U_FILE_ACCESS_ERROR with tensorflow_text


tgsmith61591:
### System information
```
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.5 LTS
Release:	16.04
Codename:	xenial
```
- **TensorFlow Serving installed from (source or binary)**: Using a stock docker image
- **TensorFlow Serving version**: `tensorflow/serving:nightly`
- **TensorFlow version**: `2.2.0rc3`
- **TensorFlow Text version**: `2.2.0.rc2`
- **Python version**: `3.6.10`

### Describe the problem

This was posted as an issue on the tensorflow/text repo (https://github.com/tensorflow/text/issues/281), but I suspect since it has to do with the way serving is built, I think it belongs here too. When serving a model that has a `BERTTokenizer` from `tf_text`, I'm getting:

```
{'error': 'U_FILE_ACCESS_ERROR: Could not retrieve ICU NFKC_CaseFold normalizer\\n\\t [[{{node transformer_network/bert_embeddings/bert-tokenizer/CaseFoldUTF8/CaseFoldUTF8}}]]'}
```

I suspect `tensorflow_text` is not successfully built as part of the `third_party` dependencies in tensorflow_serving, which is why it cannot find the op.

tgsmith61591:
For whatever it's worth, I built `tensorflow_serving` with the following changes:

1. Updated the `workspace.bzl` to point to the new tf-text distro:

  ```bazel
      http_archive(
          name = "org_tensorflow_text",
          sha256 = "468abebaa91b08fc0bd34ca7b8862ff7d7853556f231ba9b06f887c3aaf5f1bc",
          strip_prefix = "text-2.2.0-rc2",
          urls = [
              "https://github.com/tensorflow/text/archive/v2.2.0-rc2.zip",
          ],
          patches = ["@//third_party/tf_text:tftext.patch"],
          patch_args = ["-p1"],
          repo_mapping = {"@com_google_re2": "@com_googlesource_code_re2"},
      )
  ```

2. Updated the `tftext.patch` to:

  ```diff
diff -Naur a/tensorflow_text/core/kernels/BUILD b/tensorflow_text/core/kernels/BUILD
--- a/tensorflow_text/core/kernels/BUILD	2020-04-27 14:23:30.000000000 -0500
+++ b/tensorflow_text/core/kernels/BUILD	2020-04-27 15:39:35.000000000 -0500
[...]

PR #1616, opened 2020-04-28
Fix U_FILE_ACCESS_ERROR in tensorflow-text

tgsmith61591:
Addresses #1613 by reworking the `third_party/icu` build chain to more closely resemble the `tf_text/third_party/icu`. I am able to load and serve a model with the `BertTokenizer` without any problems.


tgsmith61591:
@googlebot I signed it!

tgsmith61591:
It's been a week, can someone acknowledge this, please?

elmadj:
Is there a target date for the Pull Request Review? We need to know the time frame internally so we can adjust/find other solutions.

misterpeddy:
Note, the person who originally wrote this integration is on leave and will likely be able to look at it the week after next at the earliest.

broken:
Hi @tgsmith61591
Apologies for the delay, today is my first day back and I'll take a look at this.

broken:
With TF Serving, they should be building all of the ICU data in, so there shouldn't be a need for patching. The reason for this is that TF & TF Text both have different patched versions of ICU, which works when they are separate libraries. However, TF Serving builds everything together in a single binary, so rather than having to maintain a patch file that combines the two, they've decided to build all the ICU data in (for hope to prevent issues like this).

I suspect if you copy the http_archive currently defined in tensorflow_serving/workspace.bzl to where you have it in WORKSPACE, this should fix the issue. The rest of the changes should not be necessary. I haven't tested this yet, but my reasoning is that I believe TF Serving 2.2.0-rc# is now relying on the TF version of ICU instead of its own which is probably due to an update in the version of Bazel packaged with the Docker build image.

tgsmith61591:
Thanks, @broken, I'll try that out

broken:
This looks good to me. Were you able to test?

tgsmith61591:
I rebuilt yesterday with these changes and ran into the same `U_FILE_ACCESS_ERROR` issue. Doesn't appear the CI checks are public, either

broken:
Ok; I'll build today and look closer at the artifacts.

tgsmith61591:
Thanks @broken

broken:
Okay, I found a few things.

First, the Bazel build for ICU does appear to not be including all the data. I've been having trouble figuring out the right changes needed, so I've reached out to the ICU engineers for help.

Second, it looks like the reason this wasn't caught in TF Serving's tests was that errors in inputs to tf.control_dependencies are silently swallowed when in graph mode. This is a bug in core TF that I've notified them about.

Finally, the sha256 of the ICU zip from GitHub appears to have changed. It's an easy fix, but I find it odd.

broken:
I worked with ICU at trying to bring the bazel build up to date to include all the ICU files. The number of changes is actually quite lengthy, and still requires patching build files in the correct places of the repo bazel checks out to get right. So I'm going to push for having a single patch that includes both TF & TF Text's patches.

I've also found a work around that fixes the regression model to catch this in testing.

And for the last point, it appears like github (today?) rolled back whatever changes it had done to the library that caused the sha to update.

@tgsmith61591 Since the changes should be done in parts, I'm just going to go ahead and send the changes myself to tf serving to make it easier and close the feedback loop. I have discussed with them, and this is a blocker for 2.2.0, so my changes will be included in the next release.

broken:
These changes to the build/workspace and test are in. I expect them to be in the next nightly release. I'll update the corresponding issues.

As a side note, turns out the icu release archive change was due to github previously not including some of the data binaries that were in the repo. We're switching to a Google hosted mirror to prevent this in the future.

netfs:
@broken do we still need this PR? i guess we can close it?

broken:
Yes. This can be closed.

PR #1622, opened 2020-05-09
Supports serialized POST body messages in protocol buffer format

shkwsk:
Hello.

This patch supports serialized POST body messages in protocol buffer format in the RESTful API.

Raw json message in the RESTful API is too much traffic for applications that handle a lot of requests.
Fortunately, gzip format is already supported, but you can expect further compression by serializing in protocol buffer format.

API Specification:
If "protobuf" is specified for "model_name" in the current API specification, body messages in protocol buffer format will be interpreted and predicted.
In this case, "model_name" must be specified by ModelSpec.name.

### DIFF
--- a/tensorflow_serving/model_servers/http_rest_api_handler.cc
+++ b/tensorflow_serving/model_servers/http_rest_api_handler.cc
@@ -94,24 +94,34 @@ Status HttpRestApiHandler::ProcessRequest(
   if (http_method == "POST" &&
       RE2::FullMatch(string(request_path), prediction_api_regex_, &model_name,
                      &model_version_str, &method)) {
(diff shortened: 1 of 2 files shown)


shkwsk:
It's been two weeks now, could you please confirm this?

netfs:
if you want to pass protobufs and seem to care about performance, please use gRPC.

i do not think we want to add support fo pass protobuf data via REST APIs.

netfs:
i am going to close this request, see my previous comment on using grpc if you want to pass protos around.

shkwsk:
@netfs 
Hmmm, I think it's useful in legacy environments (e.g. CentOS6) where gRPC is not available.

PR #1643, opened 2020-05-27
Fix icu dependency checksum.

christisg:


netfs:
wait for the build to pass before merging.

also this commit does not seem to be a cherry-pick of 6ad81047a to preserve history. this is OK for now, but in future, please cherry-pick upstream changes than making a new one on this branch.

PR #1687, opened 2020-07-14
Added model_platform to the model_config_list documentation

hanneshapke:
It seems the TF Serving docs for `model_config_list` is missing the required argument `model_platform`. If not present, the server will fail with this error below:

```
Error: Invalid argument: Illegal setting neither ModelServerConfig::model_type (deprecated) nor ModelServerConfig::model_platform.
```

Could the two additional lines be added to the example configuration file for the `model_config_list`?


netfs (approved):

netfs:
seems the document was updated in 084eaeb15fdc87d83b8c19f558dc1f56bd3a024e

PR #1702, opened 2020-07-29
Mnist Support Tensorflow 2.x

Hui-Zhi:
Change ```tf.app.flags.FLAGS``` to ```tf.compat.v1.app.flags.FLAGS``` so mnist example can support Tensorflow 2.x


netfs:
fixed in 06fbf878a98c8b

PR #1727, opened 2020-09-02
prevent unloading the serving versions.

jianpingzju:
In specific version policy mode, when the filesystem crashed or the version directory was deleted by mistake, the serving version may be unloaded. Which is a disaster for online service. So this pull request tries to prevent unloading the versions in configuration. If we need unload versions, changing then reloading configuration may be a good choice.


netfs:
what happens if your serving process restarts (while the underlying filesystem or directory has been deleted). that still will be a disaster. no? IOW this fix is only useful until your process restarts or is preempted and gets rescheduled elsewhere in the cluster -- and when that happens you will have failures (disaster as you say). no?

PR #1623, opened 2020-05-09
Add metrics to the Prometheus API

shkwsk:
Hello.

This patch adds latency and availability metrics for Predict method in gRPC API to the Prometheus API.
The metrics are measured for the entire process and for each model.

The added metric items are as follows.
* /tensorflow/serving/all_request_count: the total number of requests in requests processing
* /tensorflow/serving/all_request_fail_count: number of failed requests in the entire request process
* /tensorflow/serving/all_request_latency_usec: total request processing latency
* /tensorflow/serving/all_request_latency_histogram_usec: a histogram of the total latency of the request processing
* /tensorflow/serving/model_request_count: number of requests by model
* /tensorflow/serving/model_request_fail_count: number of failed requests by model
* /tensorflow/serving/model_request_latency_usec: latency by model
* /tensorflow/serving/model_request_latency_histogram_usec: latency histogram by model

PR #1765, opened 2020-11-04
Add more exported metrics for prediction

mmischitelli:
I'm adding:
- Counters for successful and failed predictions (per-model and for all, aggregated by api/runtime)
- Total inference time (aggregated by api/runtime)


netfs:
there is #1623 that adds similar changes. would getting that change in, work for you too?

mmischitelli:
> there is #1623 that adds similar changes. would getting that change in, work for you too?

Yes I think so. It has some conflicting changes I see, though

netfs:
> > there is #1623 that adds similar changes. would getting that change in, work for you too?
> 
> Yes I think so. It has some conflicting changes I see, though

great. lets close this PR, and please add your comments/needs to #1623 (that i have started reviewing) -- if you see something missing. will that be OK?

PR #1775, opened 2020-11-24
Releng 20201119 143440

shadowdragon89:


tensorflow_serving/tools/pip_package/setup.py
```diff
@@ -33,7 +33,7 @@
 # Set when releasing a new version of TensorFlow Serving (e.g. 1.0.0).
 _VERSION = '2.4.0-rc2'
 # Have this by default be open; releasing a new version will lock to TF version
-_TF_VERSION = '>=2.4,<3'
+_TF_VERSION = '>=2.4rc2,<3'
```
netfs, line 36: why change this file? we only need to revert temporary changes done to docker file.
fixing this will given an illusion that published pip package (at pypi) has the correct
requires clause (while it does not have. note, 2.4.0-rc2 pip package is broken).

shadowdragon89: That's right. I missed the part that PIP package is not re-build every time.
Reverting changes in docker file may temporarily break the 2.4 docker build. Is that OK before the final 2.4 release? Or should we make the change when the final release start.

netfs: well docker builds are broken anyways at the 2.4.0-rc2 release tag. so should be ok.

netfs (approved):

PR #1785, opened 2020-12-15
Add grpc_max_threads

zhjunqin:
Add set max gprc server threads to avoid memory increasing a lot when TF serving start and receive messages.

```
  ::grpc::ResourceQuota res_quota;
  res_quota.SetMaxThreads(server_options.grpc_max_threads);
  builder.SetResourceQuota(res_quota);
```

If there is no max threads, then Grpc server will create a lot of threads if first comming messages hang during TF lazy Initialization.

You can check Grpc server source code :

https://github.com/grpc/grpc/blob/659f5e18e4811903a113bc5aa5dd0832523763ba/src/cpp/thread_manager/thread_manager.cc#L149

```
        if (!shutdown_ && num_pollers_ < min_pollers_) {
          if (grpc_resource_user_allocate_threads(resource_user_, 1)) {
            // We can allocate a new poller thread
            num_pollers_++;
            num_threads_++;
            if (num_threads_ > max_active_threads_sofar_) {
              max_active_threads_sofar_ = num_threads_;
            }
            // Drop lock before spawning thread to avoid contention
            lock.Unlock();
            WorkerThread* worker = new WorkerThread(this);
            if (worker->created()) {
              worker->Start();
            } else {
              // Get lock again to undo changes to poller/thread counters.
              grpc_core::MutexLock failure_lock(&mu_);
              num_pollers_--;
              num_threads_--;
              resource_exhausted = true;
              delete worker;
            }
```

The most worst thing is it would lead to TF Serving docker container OOM if memory used by Grpc theads exceeds max memory limit.

### DIFF
--- a/tensorflow_serving/model_servers/server.h
+++ b/tensorflow_serving/model_servers/server.h
@@ -43,6 +43,7 @@ class Server {
     tensorflow::int32 grpc_port = 8500;
     tensorflow::string grpc_channel_arguments;
     tensorflow::string grpc_socket_path;
+    tensorflow::int32 grpc_max_threads = 4.0 * port::NumSchedulableCPUs();
 
     //
     // HTTP Server options.
--- a/tensorflow_serving/model_servers/main.cc
+++ b/tensorflow_serving/model_servers/main.cc
@@ -174,6 +174,9 @@ int main(int argc, char** argv) {
                        "A comma separated list of arguments to be passed to "
                        "the grpc server. (e.g. "
                        "grpc.max_connection_age_ms=2000)"),
+      tensorflow::Flag("grpc_max_threads",
+                       &options.grpc_max_threads,
+                       "Max grpc server threads to handle grpc messages."),
       tensorflow::Flag("enable_model_warmup", &options.enable_model_warmup,
                        "Enables model warmup, which triggers lazy "
                        "initializations (such as TF optimizations) at load "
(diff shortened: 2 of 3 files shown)


zhjunqin:
I think it can fix some cases in issue https://github.com/tensorflow/serving/issues/1215

tensorflow_serving/model_servers/server.h
```diff
@@ -43,6 +43,7 @@ class Server {
     tensorflow::int32 grpc_port = 8500;
     tensorflow::string grpc_channel_arguments;
     tensorflow::string grpc_socket_path;
+    tensorflow::int32 grpc_max_threads = 200;
```
netfs, line 46: a function of num_cpus might be more maintainable, compared to this hard-coded value.

maybe `4.0 * port::NumSchedulableCPUs()`

?

netfs (commented):
Thanks for the PR!

tensorflow_serving/model_servers/server.h
zhjunqin: OK

zhjunqin: Fixed.

netfs (changes requested):
Thanks the change looks good.

Can you please update the commit message with more details on why this tunable/flag is needed, so folks reading the commit message (later) can know the motivation. Once you do that I can approve+merge.

Thanks!

zhjunqin:
> Thanks the change looks good.
> 
> Can you please update the commit message with more details on why this tunable/flag is needed, so folks reading the commit message (later) can know the motivation. Once you do that I can approve+merge.
> 
> Thanks!

Commit message is updated, please have a check, thanks.

netfs (approved):

Issue #1764, opened 2020-11-02
Option to disable TF serving on grpc over http when TF serving starts on unix domain socket


abhinavos7a:
### Describe the problem the feature is intended to solve
There isn't a way to just run TF serving on **unix domain socket only**, as by default TF serving is started with grpc over http on port 8500. 

In multi-user system TF serving running over network have disadvantage. Programs on the same host are able to connect to TF serving. Below are some challenges to secure TF server and clients: 

- Authorization : To restrict other programs access to TF serving APIs, authorization business logic is required by TF serving.  

- TLS connection : Since same network namespace is used, client connection to TF serving needs to be encrypted so that other user application can't listen.

An amazing and simple way to get away with above 2 problems is to start TF serving with GRPC over unix domain socket. Since unix domain socket use files as namespace, both of above problems are resolved. 

`tensorflow_model_server --grpc_socket_path=grpc_unix_domain_socket --model_config_file=models.config`

The problem is that, when TF serving is started with above config, in addition to server registering with unix domain socket,   TF serving also starts listening to port 8500 (grpc)
[code](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/server.cc#L316) 
`
Running gRPC ModelServer at 0.0.0.0:8500 ...
Running gRPC ModelServer at UNIX socket grpc_unix_domain_socket
`

Although programs would use unix domain socket for communication, but since there is an additional http application (grpc overr http) accessible by other programs , authorization is required to protect TF serving APIs
So we need a way to disable TF serving on grpc over http.

### Requirements
To start TF serving only on unix domain socket
tensorflow_model_server grpc_port=0 --grpc_socket_path=grpc_unix_domain_socket --model_config_file=models.config

lilao:
@abhinavos7a : you seem to have found a way to make TF serving work on unix domain socket. Is there anything else you need from us? Thanks!

abhinavos7a:
@lilao We need an option to disable TF serving running on GRPC over http when the intention is to just start TF serving on unix domain socket.

netfs:
Can you send a PR? Happy to review it.

abhinavos7a:
@netfs https://github.com/tensorflow/serving/pull/1790

netfs:
fixed in #1790

PR #1789, opened 2020-12-28
option to disable grpc over http

abhinavos7a:
Option to disable TF serving on grpc over http when TF serving starts on unix domain socket. 

The issue is described in more detail here
https://github.com/tensorflow/serving/issues/1764


abhinavos7a:
@googlebot I signed it!

abhinavos7a:
@googlebot I fixed it.

PR #1796, opened 2021-01-13
The number of threads for loading and unloading models is set to 1

handong0123:
`uint32 num_load_threads = 0;`
`uint32 num_unload_threads = 0;`

These two parameters are 0 by default, which means that independent thread pools are not used and run in the same thread as Serving Manager.

The threads that update, load the model, and process TensorFlow Serving requests share the same thread pool, which makes it impossible to process requests when switching models.

### DIFF
--- a/tensorflow_serving/core/aspired_versions_manager.h
+++ b/tensorflow_serving/core/aspired_versions_manager.h
@@ -110,13 +110,13 @@ class AspiredVersionsManager : public Manager,
     ///
     /// If set as 0, we don't use a thread-pool, and servable loads are
     /// performed serially in the manager's main work loop.
-    uint32 num_load_threads = 0;
+    uint32 num_load_threads = 1;
 
     /// The number of threads in the thread-pool used to unload servables.
     ///
     /// If set as 0, we don't use a thread-pool, and servable unloads are
     /// performed serially in the manager's main work loop.
-    uint32 num_unload_threads = 0;
+    uint32 num_unload_threads = 1;
 
     /// Maximum number of times we retry loading a servable, after the first
     /// failure, before we give up.
--- a/tensorflow_serving/core/aspired_versions_manager.h
+++ b/tensorflow_serving/core/aspired_versions_manager.h
@@ -110,13 +110,13 @@ class AspiredVersionsManager : public Manager,
     ///
     /// If set as 0, we don't use a thread-pool, and servable loads are
     /// performed serially in the manager's main work loop.
-    uint32 num_load_threads = 1;
+    uint32 num_load_threads = 0;
 
     /// The number of threads in the thread-pool used to unload servables.
     ///
     /// If set as 0, we don't use a thread-pool, and servable unloads are
     /// performed serially in the manager's main work loop.
-    uint32 num_unload_threads = 1;
+    uint32 num_unload_threads = 0;
 
     /// Maximum number of times we retry loading a servable, after the first
     /// failure, before we give up.
--- a/tensorflow_serving/model_servers/main.cc
+++ b/tensorflow_serving/model_servers/main.cc
@@ -105,6 +105,18 @@ int main(int argc, char** argv) {
       tensorflow::Flag("model_base_path", &options.model_base_path,
                        "path to export (ignored if --model_config_file flag "
                        "is set, otherwise required)"),
+      tensorflow::Flag("num_load_threads", &options.num_load_threads,
+                       "The number of threads in the thread-pool used to load servables."
+                       "If set as 0, we don't use a thread-pool, and servable loads are "
+                       "performed serially in the manager's main work loop, "
+                       "may casue the Serving request to be delayed. "
+                       "Default: 0"),
+      tensorflow::Flag("num_unload_threads", &options.num_unload_threads,
+                       "The number of threads in the thread-pool used to unload servables."
+                       "If set as 0, we don't use a thread-pool, and servable loads are "
+                       "performed serially in the manager's main work loop, "
+                       "may casue the Serving request to be delayed. "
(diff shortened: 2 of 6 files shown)


netfs:
> The threads that update, load the model, and process TensorFlow Serving requests share the same thread pool, which makes it impossible to process requests when switching models.

i can see the value of using different threadpools for isolation, but cannot see why it becomes "impossible" to process requests while model is loading (if that is what you mean). i'd expect >1 threads in the common pool, that will allow to serve requests and load model, yes at times causing serving requests to starve or slow down.

but "impossible" sounds a bit odd. care to explain more?

thanks!

handong0123:
> > The threads that update, load the model, and process TensorFlow Serving requests share the same thread pool, which makes it impossible to process requests when switching models.
> 
> i can see the value of using different threadpools for isolation, but cannot see why it becomes "impossible" to process requests while model is loading (if that is what you mean). i'd expect >1 threads in the common pool, that will allow to serve requests and load model, yes at times causing serving requests to starve or slow down.
> 
> but "impossible" sounds a bit odd. care to explain more?
> 
> thanks!

Maybe my description is inaccurate, The core operation of model loading is RestoreOp, which includes operations such as reading model files from storage, allocating memory, and finding the corresponding Variable, which are executed by calling the run method of Session. By default, all Session operations in a process use the same thread pool. Therefore, the loading operation and the operation processing the Serving request in the model loading process use the same thread pool, which causes the Serving request to be delayed.So there will be glitches when the model is switched.

netfs:
> > > The threads that update, load the model, and process TensorFlow Serving requests share the same thread pool, which makes it impossible to process requests when switching models.
> > 
> > 
> > i can see the value of using different threadpools for isolation, but cannot see why it becomes "impossible" to process requests while model is loading (if that is what you mean). i'd expect >1 threads in the common pool, that will allow to serve requests and load model, yes at times causing serving requests to starve or slow down.
> > but "impossible" sounds a bit odd. care to explain more?
> > thanks!
> 
> Maybe my description is inaccurate, The core operation of model loading is RestoreOp, which includes operations such as reading model files from storage, allocating memory, and finding the corresponding Variable, which are executed by calling the run method of Session. By default, all Session operations in a process use the same thread pool. Therefore, the loading operation and the operation processing the Serving request in the model loading process use the same thread pool, which causes the Serving request to be delayed.So there will be glitches when the model is switched.

Thanks. That helped clarify. Yes things can get delayed or starve due to having one threadpool for loading+serving.

tensorflow_serving/core/aspired_versions_manager.h
```diff
@@ -110,13 +110,13 @@ class AspiredVersionsManager : public Manager,
     ///
     /// If set as 0, we don't use a thread-pool, and servable loads are
     /// performed serially in the manager's main work loop.
-    uint32 num_load_threads = 0;
+    uint32 num_load_threads = 1;
```
netfs, line 114: please do not change default values. instead add two new flags to the server (`num_load_threads` and `num_unload_threads`), so people can (optionally) tweak the settings as needed for their use case.

as an example, copy/follow the pattern on how `--max_num_load_retries` flag is implemented/used to set appropriate lower level options.

https://github.com/tensorflow/serving/blob/0acaa63b45a5c461dcb4808f8745ff9088a0f6e6/tensorflow_serving/model_servers/main.cc#L108

you can add new flags around this.

handong0123: ok, I had added two new flags to the server (num_load_threads and num_unload_threads).

tensorflow_serving/core/basic_manager.cc
```diff
@@ -230,8 +230,8 @@ Status BasicManager::Create(Options options,
   return Status::OK();
 }
 
-BasicManager::BasicManager(Env* const env, const uint32 num_load_threads,
```
netfs, line 233: why do we need this change?

handong0123: I don’t quite understand why max_num_load_retries is not modified by const, so in order to be consistent, I made changes, which may not be appropriate.

netfs: i'd recommend reverting this file, as i dont see it related to the change you are doing. can you please revert this file?

also can you please update the commit message to read better -- mention the new flags that you have added, and what they do.

thanks!

handong0123: ok, I've completed.

netfs:
thanks for the change and your contribution. will approve and merge shortly.

tensorflow_serving/model_servers/main.cc
```diff
@@ -105,6 +105,18 @@ int main(int argc, char** argv) {
       tensorflow::Flag("model_base_path", &options.model_base_path,
                        "path to export (ignored if --model_config_file flag "
                        "is set, otherwise required)"),
+      tensorflow::Flag("num_load_threads", &options.num_load_threads,
+                       "The number of threads in the thread-pool used to load servables."
+                       "If set as 0, we don't use a thread-pool, and servable loads are "
+                       "performed serially in the manager's main work loop, "
+                       "may casue the Serving request to be delayed. "
+                       "Default: 0"),`
```
netfs, line 113: there is a stray ``` at the end. please remove. this is i think causing the tests to break:

https://source.cloud.google.com/results/invocations/dc6486cd-9c79-486b-b5db-0632da419481/targets

can you also ensure all the tests pass?

thanks!

handong0123: This is my mistake, now corrected

netfs (approved):

netfs:
Thanks!

PR #1801, opened 2021-01-22
Add reconfig poll patch

### LINKED ISSUE #1519: Aspired servables/versions aren't refreshed following config updates
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04, RHEL 8
- **TensorFlow Serving installed from (source or binary)**: Both
- **TensorFlow Serving version**: 1.15.0

### Describe the problem
Currently when the configured model list is updated via a call to `handleReloadConfigRequest`, the request thread blocks until any newly added models become available.

Their availability however depends on the filesystem polling thread rescanning the filesystem at some periodic interval, meaning that there's an arbitrary delay before the requested changes actually take effect and the RPC returns.

This problem may not be very noticeable with the default polling interval of 1 second, but seems undesirable for longer intervals and in particular makes API-based dynamic reconfiguration incompatible with the `--file_system_poll_wait_seconds=0` setting (in this case all `handleReloadConfigRequest` calls time-out and do not take effect).

### Exact Steps to Reproduce
1. Start `tensorflow_model_server` with the `--file_system_poll_wait_seconds=0` option and empty initial config (no models)
2. Call `handleReloadConfigRequest` API with a `ModelListConfig` containing a (valid) new model. It will hang indefinitely or until the grpc deadline

I have opened PR #1518 with proposed fix.

gkumbhat:
This patch is essentially porting the patch submitted by @njhill  in PR #1518 to `2.4.0` version

Currently when the configured model list is updated via a call to handleReloadConfigRequest, the request thread blocks until any newly added models become available.

Their availability however depends on the filesystem polling thread rescanning the filesystem at some periodic interval, meaning that there's an arbitrary delay before the requested changes actually take effect and the RPC returns.

This problem may not be very noticeable with the default polling interval of 1 second, but seems undesirable for longer intervals and in particular makes API-based dynamic reconfiguration incompatible with the --file_system_poll_wait_seconds=0 setting (in this case all handleReloadConfigRequest calls time-out and do not take effect).

Fixes #1519

### DIFF
--- a/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
+++ b/tensorflow_serving/sources/storage_path/file_system_storage_path_source.cc
@@ -293,9 +293,9 @@ Status PollFileSystemForConfig(
 
 // Determines if, for any servables in 'config', the file system doesn't
 // currently contain at least one version under its base path.
-Status FailIfZeroVersions(const FileSystemStoragePathSourceConfig& config) {
-  std::map<string, std::vector<ServableData<StoragePath>>>
-      versions_by_servable_name;
+Status FailIfZeroVersions(const FileSystemStoragePathSourceConfig& config,
+  std::map<string, std::vector<ServableData<StoragePath>>>&
+      versions_by_servable_name) {
   TF_RETURN_IF_ERROR(
       PollFileSystemForConfig(config, &versions_by_servable_name));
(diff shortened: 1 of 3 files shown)


christisg (commented):
Thanks for the contribution!

PR #1808, opened 2021-02-10
Specifying compute capabilities in the TF 2.2 .bazelrc file

rhaertel80:
This reduces the startup time on GPUs and is in line with TF 2.3 changes


netfs (approved):
Thanks for the change!

PR #1847, opened 2021-04-23
_TF_REQ is not compatible `tensorflow-cpu` requirements using pip

kimdwkimdw:
When `tensorflow-cpu` is already installed, `tensorflow-serving-api` package cannot detect tesnroflow installation.

`https://pypi.org/project/tensorflow-serving-api/` should follow `tensorflow-cpu`

`https://pypi.org/project/tensorflow-serving-api-gpu/` is now folloing `tensorflow-gpu` requirement.

### DIFF
--- a/tensorflow_serving/tools/pip_package/setup.py
+++ b/tensorflow_serving/tools/pip_package/setup.py
@@ -44,7 +44,7 @@ if '--project_name' in sys.argv:
   sys.argv.remove('--project_name')
   sys.argv.pop(project_name_idx)
 
-_TF_REQ = ['tensorflow'+_TF_VERSION_SANITIZED]
+_TF_REQ = ['tensorflow-cpu'+_TF_VERSION_SANITIZED]
 
 # GPU build (note: the only difference is we depend on tensorflow-gpu so
 # pip doesn't overwrite it with the CPU build)


kimdwkimdw:
@googlebot I signed it!

johnperez416 (approved):

tensorflow_serving/tools/pip_package/setup.py
```diff
@@ -44,7 +44,7 @@
   sys.argv.remove('--project_name')
   sys.argv.pop(project_name_idx)
 
-_TF_REQ = ['tensorflow'+_TF_VERSION_SANITIZED]
+_TF_REQ = ['tensorflow-cpu'+_TF_VERSION_SANITIZED]
```
netfs, line 48: this will break for people who have `tensorflow` package installed:

https://pypi.org/project/tensorflow/

is there a way to specify a list of packages, where any one of them can be installed to satisfy the requirement?

kimdwkimdw: How about this?

https://github.com/tensorflow/serving/pull/1847/commits/f183188d7d555e88e111c1360d135f2490e6ac85

```
import pkg_resources

pkg_released = set(["tensorflow", "tensorflow-cpu"])
pkg_installed = set(map(lambda pkg: pkg.project_name, pkg_resources.working_set))
tensorflow_installed = list(pkg_released & pkg_installed) + ["tensorflow"]

_TF_REQ = [tensorflow_installed[0]+_TF_VERSION_SANITIZED]
```

tensorflow_serving/tools/pip_package/setup.py
```diff
@@ -44,7 +45,11 @@
   sys.argv.remove('--project_name')
   sys.argv.pop(project_name_idx)
 
-_TF_REQ = ['tensorflow'+_TF_VERSION_SANITIZED]
+pkg_released = set(["tensorflow", "tensorflow-cpu"])
+pkg_installed = set(map(lambda pkg: pkg.project_name, pkg_resources.working_set))
```
netfs, line 49: for my understanding, is pkg_resources.working_set from the machine where this package is going to be installed OR is this from the machine where the package is being built? we obviously want the former (list packages installed on the target).

kimdwkimdw: https://setuptools.readthedocs.io/en/latest/pkg_resources.html#workingset-objects
```
working_set object is initialized from sys.path when pkg_resources is first imported
```

Yes. the former is right.

kimdwkimdw: It was my misunderstanding. That is for the latter, not the former one.

I think there is no solution for `conditional requirements`.

Same as `tensorflow` package,  https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/setup.py

Creating new package is another solution with `tensorflow-serving-api-cpu`
And then add `tensorflow-cpu` to `_TF_REQ` when package name is `tensorflow-serving-api-cpu`
ex) https://pypi.org/project/tensorflow-serving-api-cpu/

PR #1912, opened 2021-09-08
Copy old client to new package and updated unit tests

jeremiahmao:
Copied the old client to the new “test_client” package and updated all the unit tests that use this client. Also added 'Test' Prefix to all classes and structs. All this is done to clear space for the development of a new public http client API. @wenbozhu will help upgrade all existing google-internal call sites of the old client. After that, I will delete everything directly under net_http/client.

### DIFF
--- a/tensorflow_serving/util/net_http/client/test_client/internal/BUILD
+++ b/tensorflow_serving/util/net_http/client/test_client/internal/BUILD
@@ -1,0 +1,27 @@
+# Description: a lightweight http client
+
+package(
+    default_visibility = [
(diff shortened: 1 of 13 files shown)


jeremiahmao:
@googlebot I signed it!

wenbozhu (dismissed):

netfs:
can you please rewrite the commit message to be more readable.

something like this may be better:

===
Prepare for development of new http client API.

Copy the old client to the new “test_client” package and update affected unit-tests.
There is no functional change to the existing API or interfaces.
===

netfs (approved):

Issue #2164, opened 2023-07-27
Display full message in GRPC exception log


Prabha-Veerubhotla:
## Feature Request
If this is a feature request, please fill out the following form in full:

### Describe the problem the feature is intended to solve
While using Tensorflow serving,  the exception message log is TRUNCATED. 
Ex:
```
Caused by: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: xxxx...TRUNCATED
```

### Describe the solution
Full logs should be displayed without any truncation. 

### Additional context
The Tensorflow serving client side config for max inbound message size 
`.maxInboundMessageSize()`
is set to
`int32max`  to match with server side config 
https://github.com/tensorflow/serving/blob/6a9d0fdafe326605cad1cae60dea0dd165bd2bb4/tensorflow_serving/model_servers/server.cc#L385

The origination code seems to be the following as per stack trace:
https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/ClientCalls.java#L275

### System information
- **TensorFlow Serving version**: 2.11.0.5

### Source code / logs
Prediction in Tensorflow serving:
https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/prediction_service.proto#L23

singhniraj08:
@Prabha-Veerubhotla,

TF serving binary has C++ dependencies. C++ logging has a hard coded limit on log size 15K which results in this truncation of log messages. A typical workaround of this problem is to output every line as a separate record. This approach can rapidly increase memory usage which can result in crash due to a memory exceeded error. 
Please let us know, if this is completely blocking you, we can try looking for alternatives. Thank you!

singhniraj08:
@Prabha-Veerubhotla,

Going through complete list of available [command line flags](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/main.cc#L81), I couldn't find any flag or param to set custom error message limit. Thanks.

Prabha-Veerubhotla:
@singhniraj08 how about adding a new command line argument to set the custom error message limit [here](https://github.com/tensorflow/serving/blob/701e5c049f71aadb48463908b86be51f0171c0dd/tensorflow_serving/model_servers/grpc_status_util.cc#L25).

I am currently blocked with this as I cannot look at the complete error message. There are some features that are failing in TF serving.

singhniraj08:
@Prabha-Veerubhotla,

[...]

PR #2185, opened 2023-09-21
Log message without truncation while leaving grpc limits as is

asamadiya:
Truncated messages as described in https://github.com/tensorflow/serving/issues/2164 kills debug-ability. Rather than expecting the client to receive potentially large messages, we simply log the full message on the server.

Testing: WIP


asamadiya:
SIgned.

PR #4138, opened 2026-05-28
Fix stack exhaustion DoS in json parsing via iterative shape traversal

deezsecc:
## Description

This PR fixes a **stack exhaustion / Denial of Service (DoS)** vulnerability in the TensorFlow Serving REST API JSON parsing logic.

---

## Root Cause

When parsing incoming model prediction requests, `rapidjson` processes the request document iteratively. However, the post-parsing step in `GetDenseTensorShape()` (inside `tensorflow_serving/util/json_tensor.cc`) recursively traversed nested arrays (`val[0]`) to compute the tensor's shape **without enforcing a recursion depth guard or limit**.

An unauthenticated attacker could send a payload with a deeply nested array structure (e.g., 50,000 nested layers like `[[[[...` `[1]` `...]]]]`), causing a stack overflow (`SIGSEGV`) and instantly crashing the `tensorflow_model_server` process.

A secondary recursive execution vector also existed in `FillTensorProto()`, which recursively parses nested elements up to the tensor's rank without validating that the rank is within safe limits.

---

## Fix

This patch mitigates the vulnerability on both vectors:

| Vector | Fix |
|---|---|
| **Iterative Traversal** | Refactored `GetDenseTensorShape()` to be fully flat and iterative, using a loop and pointer to resolve shape — eliminating recursion at the shape-calculation level |
| **Rank Capping** | Added a hard limit (`kMaxTensorRank = 254`, aligned with TensorFlow's `TensorShape::MaxDimensions()`) inside `GetDenseTensorShape()` to stop adding dimensions beyond safe nesting depth |
| **Early Rejection** | Added an early check in `FillTensorProto()` to reject payloads with rank > 254 via `errors::InvalidArgument`, preventing deep recursion on the value-filling path |

---

## Verification

- [x] Iterative `while` loop behaves identically to the original recursion on standard payloads
- [x] Deeply nested structures are capped early and rejected via `FillTensorProto` propagating `InvalidArgument` instead of exhausting the thread stack

### DIFF
--- a/tensorflow_serving/util/json_tensor.cc
+++ b/tensorflow_serving/util/json_tensor.cc
@@ -346,13 +346,25 @@ Status AddValueToTensor(const rapidjson::Value& val, DataType dtype,
 // `val` can be scalar or list or list of lists with arbitrary nesting. If a
 // scalar (non array) is passed, we do not add dimension info to shape (as
 // scalars do not have a dimension).
-void GetDenseTensorShape(const rapidjson::Value& val, TensorShapeProto* shape) {
-  if (!val.IsArray()) return;
-  const auto size = val.Size();
-  shape->add_dim()->set_size(size);
-  if (size > 0) {
-    GetDenseTensorShape(val[0], shape);
+constexpr int kMaxTensorRank = 254;
+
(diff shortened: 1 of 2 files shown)


deezsecc:
@ch1nlu @rtg0795 Please review the PR. Thank you!

